PR #14, opened 2023-01-25
[ARC-532] Enable Pylint and fix issues

SamTheisens:
branch: feature/ARC-532-introduce-pylint


ewaldschrader (commented):
Walked through it with Sam. 

TO DO: 
add comment for overrides
create failure scenario when k8s parsing failed
v1_probe:  add type

PR #33, opened 2023-02-18
[ARC-579] Add custom (rich) logger

SamTheisens:
branch: feature/ARC-579-improve-logging-readability


ewaldschrader (approved):

ewaldschrader:
I tried testing this branch, but I'm not seeing any changes to the logging format when I run the example project. Am I missing something?

SamTheisens:
> I tried testing this branch, but I'm not seeing any changes to the logging format when I run the example project. Am I missing something?

Are you running them through the dagit UI? Indeed there is no change visible there. But if you run via `pipenv run run` (like in Jenkins) you should be seeing the difference.

PR #40, opened 2023-03-02
[ARC-584] Conditionally load mpyl from lib or sources

SamTheisens:
depending on whether we run a local or ci build

branch: feature/ARC-584-clean-up-mpyl-example-and-try-to-use-it-as-the-main-runner


ewaldschrader (commented):
Looks good to me. One thing: when doing a local & dryrun, it still does the deploy to kubernetes step. I guess we should change that.

SamTheisens:
> Looks good to me. One thing: when doing a local & dryrun, it still does the deploy to kubernetes step. I guess we should change that.

I just check from my machine (`pipenv run run`) and I can't reproduce. Could it be that it _looks_ like it's doing a deploy? You will see a `yaml` dump during the dry run

PR #42, opened 2023-03-02
[ARC-524] Add Step to deploy job to K8s

ewaldschrader:
branch: feature/ARC-524-K8s-deploy-job


src/mpyl/steps/deploy/k8s/chartbuilder.py
```diff
@@ -249,10 +249,14 @@ def to_ingress_routes(self) -> Optional[V1AlphaIngressRoute]:
...
 
 class JobChartBuilder(ChartBuilder):
-    def to_chart(self, ) -> dict[str, str]:
+    def to_chart(self) -> dict[str, str]:
         chart = {'deployment': to_yaml(self.to_deployment()),
-                 'serviceaccount': to_yaml(self.to_service_account()),
-                 'job': to_yaml(self.to_job())}
+                 'serviceaccount': to_yaml(self.to_service_account())}
+
+        if self.step_input.project.kubernetes.cron:
```
koeves, line 256: Maybe it would be nicer to create a `Kubernetes CronJob Deploy` step? Then we can have separate job and cronjob builders

tests/steps/deploy/k8s/test_chart_builder.py
```diff
@@ -80,6 +81,13 @@ def test_should_validate_against_crd_schema():
...
+def test_service_chart_roundtrip(template):
+    charts = _build_chart(ServiceChartBuilder())
     _roundtrip(template_path, template, charts)
+
+
+@pytest.mark.parametrize('a',
+                         ['configmap', 'cronjob', 'scheduledsparkapplication', 'sealedsecret', 'serviceaccount'])
+def test_job_chart_roundtrip(a):
+    charts = _build_chart(JobChartBuilder())
+    _roundtrip(job_template_path, a, charts)
```
koeves, line 93: This test is still failing because of a key error. Probably related to how these yamls were generated, which I am unsure of if I did the correct thing

tests/stages/test_discovery.py
```diff
@@ -20,7 +20,8 @@ def test_should_find_invalidated_test_dependencies(self):
         assert len(
             find_invalidated_projects_for_stage(projects, Stage.BUILD, [History(0, "revision", touched_files)])) == 1
         assert len(
-            find_invalidated_projects_for_stage(projects, Stage.TEST, [History(0, "revision", touched_files)])) == 2
+            find_invalidated_projects_for_stage(projects, Stage.TEST, [
+                History(0, "revision", touched_files)])) == 2  # TODO: the tests say it's one, not sure what's changed
```
koeves, line 24: This is equal to 1 now, was like that before my changes as well. Not sure what changed here

src/mpyl/steps/deploy/k8s/chartbuilder.py
SamTheisens: We could do it that way. Not sure which is cleaner. If we create separate jobs there's a possibility to misconfigure i.e. a `Deploy Cron Job` without the cron field set for example.

PR #51, opened 2023-03-14
feature/ARC 524 K8s deploy job fixup

SamTheisens:
- Reusable logic between job and service is done via extracted method  (favor composition over inheritance). The same goes for the chart generation.
 - A single job type doing both cron and normal job. Curious to know if you think that’s the right way to go
Included standard k8s crd schema into project.yml I tried referring to the original schema via url, but with overrides. That didn’t work because json schema v7 doesn’t support such things. I tried a later version but still didn’t get it to work. So I ended up copy/pasting.
 - Using standard k8s model objects so we don’t need to create our own DTO https://github.com/Vandebron/mpyl/commit/d9f209fe030f701d868c9d1671a3706e9b88a672#diff-f2e93ed09e3ea127c21228d5487fed442fe1b3364117f6a7ec4705cd660a72a0R151-R157


koeves (approved):
Nice improvements! It's fine to do jobs and cronjobs within one step I think. The code is not complicated, and there are only these two types of jobs that we have..

PR #60, opened 2023-03-17
Feature/arc 588 fix master pipeline

koeves:
Monorepo counterpart: https://github.com/Vandebron/Vandebron/pull/9995


src/mpyl/steps/models.py
```diff
@@ -87,7 +87,7 @@ def from_configuration(run_properties: Dict, config: Dict):
 
         versioning = VersioningProperties(revision=versioning_config['revision'],
                                           branch=versioning_config['branch'],
-                                          pr_number=int(pr_num) if pr_num.isdigit() else None,
```
koeves: I thought this was needed because if the pr num env var isn't set, it's parsed as the string 'N/A'.. Is that changed now?

SamTheisens: Yes it is, the problem was that in the monorepo, the default `pyaml_env` was used instead of our own
```from mpyl.utilities.pyaml_env import parse_config```
Ours replaces `N/A` with `None` so we can have optional config properties.

koeves: Ah nice, good to know!

koeves (commented):
LGTM (can't accept my own PR)

ewaldschrader (approved):

PR #64, opened 2023-03-21
[ARC-610] Switch notification channel to project-mpyl-notifications

Jorg88:
branch: feature/ARC-610-switch-notification-channel


koeves (commented):
Looks good! Provided we confirm what's up with the lock file

PR #62, opened 2023-03-20
Add run_properties schema

ewaldschrader:
branch:

PR #67, opened 2023-03-28
[ARC-595] Support nested job

SamTheisens:
- Automatically resolve target specific fields in dictionary
 - Allow for `None` in yamls
 - Update `project.yml` schema with changes to `job`
 - Correct hash for local changes

branch: feature/ARC-595-cli-improvements

PR #68, opened 2023-03-28
fix build parameters comments

ewaldschrader:
fixed comments from this PR: 
https://github.com/Vandebron/mpyl/pull/62

### DIFF
--- a/src/mpyl/schema/run_properties.schema.yml
+++ b/src/mpyl/schema/run_properties.schema.yml
@@ -44,7 +44,7 @@ properties:
             type: string
           user_email:
             description: Email of of the user that triggered the run
-            type: ["string", "null"]
+            type: string
             format: idn-email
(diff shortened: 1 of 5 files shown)


src/mpyl/schema/run_properties.schema.yml
```diff
@@ -44,7 +44,7 @@ properties:
             type: string
           user_email:
             description: Email of of the user that triggered the run
-            type: ["string", "null"]
+            type: string
```
SamTheisens, line 47: Actually null, is good here. But it only _really_ works when this change is merged https://github.com/Vandebron/mpyl/pull/67/commits/0a56741d628df881e5b02df48306a3319fc5362c

SamTheisens (approved):
Looks good. You may revert the `string`|`null` union type change

PR #72, opened 2023-04-06
[ARC-613] Image security scan

ewaldschrader:
branch: feature/ARC-613-image-scan-trivy

- add scan stage
- echo scan step
- add docker scan step


src/mpyl/steps/__init__.py
```diff
@@ -105,13 +105,13 @@ class Step(metaclass=IPluginRegistry):
     """Is set to something other than `ArtifactType.NONE` if this step depends on an artifact produced by the execution
     of an earlier step. For example: a step in the `Deploy` stage, may need to deploy a docker image that was produced
     in the `Build` stage."""
-    before: Optional[Step]
-    after: Optional[Step]
+    before: Optional[list[Step]]
+    after: Optional[list[Step]]
```
koeves, line 109: Wondering if an optional list makes sense.. Is there a difference between a `None` and an empty list? If there're no steps before or after, should be an empty list I think

src/mpyl/steps/steps.py
```diff
@@ -125,19 +125,22 @@ def _execute_stage(self, stage: Stage, project: Project, dry_run: bool = False)
                 artifact: Optional[Artifact] = self._find_required_artifact(project, executor.required_artifact)
                 result = Output(success=True, message='')
                 if executor.before:
-                    result = self._execute(executor.before, project, self._properties,
-                                           self._find_required_artifact(project, executor.before.required_artifact),
-                                           dry_run)
+                    for before_step in executor.before:
+                        result = self._execute(before_step, project, self._properties,
+                                               self._find_required_artifact(project, before_step.required_artifact),
+                                               dry_run)
```
koeves, line 131: We should probably combine the results here then. Right now it just takes the last execution

src/mpyl/steps/test/dockertest.py
```diff
@@ -21,8 +22,8 @@ def __init__(self, logger: Logger) -> None:
             logger=logger, meta=meta,
             produced_artifact=ArtifactType.JUNIT_TESTS,
             required_artifact=ArtifactType.NONE,
-            before=IntegrationTestBefore(logger),
-            after=IntegrationTestAfter(logger)
+            before=[IntegrationTestBefore(logger)],
+            after=[ScanDocker(logger), IntegrationTestAfter(logger)]
```
koeves, line 26: Can we do any assertion that this after step succeeded? Or it triggers automatically for the sample projects and that's enough?

koeves (changes requested):
Some smaller things to check

src/mpyl/steps/__init__.py
ewaldschrader: Discussed with Sam, will revert this and add a `Scan Stage` instead.

PR #73, opened 2023-04-07
Feature/tech 271 run linting before build

Jorg88:


koeves (approved):

SamTheisens (commented):
The code looks good. But I'm doubting whether it's wise to do the validation at this point. In principle we don't want invalid `project.yml` ever to end up in master. But _if_ it still does for whatever reason, this would render the whole pipeline useless, even for local executions. Even when a single project can't be loaded. That's perhaps too much on a monerepo.

We could also do an `mpyl projects lint` at the start of the pipeline. This is an early warning, but it doesn't break everything for everyone.

Jorg88:
> The code looks good. But I'm doubting 🤔 whether it's wise to do the validation at this point. In principle we don't want invalid `project.yml` ever to end up in master. But _if_ it still does for whatever reason, this would render the whole pipeline useless, even for local executions. Even when a single project can't be loaded. That's perhaps too much on a monerepo.
> 
> We could also do an `mpyl projects lint` at the start of the pipeline. This is an early warning, but it doesn't break everything for everyone.

Yeah true I guess, but that means we completely wasted our time 😬 I think it would be good if you added a description to our tickets @SamTheisens (maybe together with @koeves) because we kinda interpreted the title wrongly and thus went down the wrong path.

PR #80, opened 2023-04-18
Feature/tech 270 interpolate pr params

Jorg88:


src/mpyl/project.py
```diff
@@ -269,6 +271,13 @@ def from_config(values: dict):
         props = values.get('properties')
         kubernetes = values.get('kubernetes')
         traefik = values.get('traefik')
+        pr_number = os.getenv('CHANGE_ID')
+
+        if traefik and pr_number:
+            for hosts in traefik['hosts']:
+                if 'pr' in hosts['host']:
+                    hosts['host']['pr'] = re.sub('{PR-NUMBER}', pr_number, hosts['host']['pr'])
+
```
koeves, line 280: We should also replace the env vars which contain {namespace} in Properties.from_config(props) with `namespace`. See e.g. https://github.com/Vandebron/Vandebron/blob/6a0a8750830417f9ba02a177716ff98342bf3057/energy-trading/trades/retrieve-market-results-job/deployment/project.yml for a project yaml with template {namespace} urls

tests/test_resources/test_invalid_namespace.yml
```diff
@@ -0,0 +1,12 @@
...
+stages:
+  build: Docker Build
+  test: Docker Test
+  deploy: Kubernetes Deploy
+maintainer: [ 'Marketplace' ]
+deployment:
+  properties:
+    env:
+      - key: WITH_NAMESPACE
+        all: "prometheus-gateway.{namespace}.svc.cluster.local"
```
koeves, line 12: In accordance with https://vandebron.atlassian.net/browse/TECH-302, I think let's get rid of vdb specific URLs and names in this PR as well

koeves (approved):
Looks good 👍

tests/test_resources/test_invalid_namespace.yml
Jorg88: I agree, but isn't this a pretty neutral/common/normal url?

koeves: Yeah okay fair enough. The test_project.yml will be modified I guess in the other card, so then nothing to do here

PR #83, opened 2023-04-26
Feature/tech 307 hide k8s details

Jorg88:


src/mpyl/steps/deploy/k8s/rancher.py
```diff
@@ -2,32 +2,40 @@
 
 from dataclasses import dataclass
 
+from ....steps import Input
 from ....project import Target
 
 
-@dataclass(frozen=True)
```
koeves, line 8: Why is it not frozen now?

src/mpyl/steps/deploy/k8s/rancher.py
```diff
@@ -2,32 +2,40 @@
...
+@dataclass()
 class ClusterConfig:
     project_id: str
     cluster_id: str
     cluster_env: str
     context: str
 
+    def __init__(self, config: dict):
```
koeves, line 16: No need to create an __init__ in a dataclass

koeves (commented):
Looks good just small comments

src/mpyl/steps/deploy/k8s/rancher.py
Jorg88: See thread in other comment

src/mpyl/steps/deploy/k8s/rancher.py
Jorg88: Yeah I wasn't sure about this part, how would you apply the logic that's currently in the `init` then? This is more clean than having to pass `cluster_configs['test'].projectId, cluster_configs['test'].clusterId, etc` for all 3 environments.

koeves: hmm yeah fair enough.. I think in other places we do this as well, and call the method `from_config()` or so

koeves: And it can be frozen as well

koeves (approved):

PR #86, opened 2023-05-05
Add VPP to schema

peterrutg:


src/mpyl/schema/project.schema.yml
```diff
@@ -32,6 +32,7 @@ properties:
         - Energy Trading
         - Big Data
         - Flex IoT
```
SamTheisens, line 34: Does `Flex IoT` still exist, or should it be removed?

peterrutg: Both `Big Data` and `Flex IoT` can be removed

PR #87, opened 2023-05-05
[TECH-392] Handle project.yml without env

peterrutg:


SamTheisens:
This PR solves the problem in a more fundamental way: https://github.com/Vandebron/mpyl/pull/88

PR #88, opened 2023-05-08
fix/TECH 270 lazy interpolate pr params

SamTheisens:
- Delay pr and namespace interpolation to deploy stage
- Allow env vars to be set via .env, for local runs
- Pull Vandebron specific config from separate, internal repo https://github.com/Vandebron/mpyl_config


Jenkinsfile
```diff
@@ -22,13 +22,19 @@ pipeline {
             }
             steps {
                 script {
+                    def gitconfig = scm.userRemoteConfigs.getAt(0)
+                    git(branch: 'main',credentialsId: gitconfig.getCredentialsId(), url: 'https://github.com/Vandebron/mpyl_config.git')
```
Jorg88, line 26: Do we still want to discuss alternatives later? I agree with doing this for now since there is some pressure from Arno for a quick solution

src/mpyl/project.py
```diff
@@ -270,27 +267,9 @@ class Deployment:
...
-                            raise KeyError(
-                                'Found "{namespace}" placeholder but no deployment.namespace was set in project.yml'
-                            )
-
-                        env_prop[key] = re.sub('{namespace}', namespace, env_prop[key])
-
-        if traefik and pr_number:
-            for hosts in traefik['hosts']:
-                if 'pr' in hosts['host']:
-                    hosts['host']['pr'] = re.sub('{PR-NUMBER}', pr_number, hosts['host']['pr'])
```
Jorg88, line 291: Could you send me and Daniel a message on Slack explaining why you solved this differently? Since we were convinced this was the best option 😬 would be a good learning opportunity.

tests/test_project.py
```diff
@@ -49,13 +49,3 @@ def test_schema_load_validation(self):
     def test_target_by_value(self):
         target = Target(Target.PULL_REQUEST)
         assert target == Target.PULL_REQUEST
-
-    def test_placeholder_replacement(self):
```
Jorg88, line 53: You don't think we need a test case like this?

Jorg88 (approved):

ewaldschrader (approved):

Jenkinsfile
SamTheisens: Sure, we can always discuss alternatives.

tests/test_project.py
SamTheisens: It has become part of the general chart composition logic now and with that it's tested by the chart roundtrip tests https://github.com/Vandebron/mpyl/pull/88/files#diff-929bc8ca4b62bd7c9555d2c69442c95034a1de4e0767816625849eed8da9d1d2L18

PR #94, opened 2023-05-09
[TECH-293] Remove hardcoded stages

ewaldschrader:
branch: feature/TECH-293-remove-hardcoded-stage


tests/steps/test_steps.py
```diff
@@ -78,21 +78,11 @@ def test_should_return_error_if_stage_not_defined(self):
...
-        assert result.output.message == "Executor 'Unknown Build' for 'build' not known or registered"
-
     def test_should_succeed_if_stage_is_not_known(self):
         project = test_data.get_project_with_stages({'test': 'Some Test'})
         result = self.executor.execute(stage=Stage.BUILD, project=project)
         assert not result.output.success
         assert result.output.message == "Stage 'build' not defined on project 'test'"
+
+    def test_find_all_steps(self):
+        assert list(Step.get_subclasses()) != ""
```
koeves, line 88: We should either assert that the list contains all of the subclasses we have (but that would mean we need to add every new one as well to this test)
Or at least we should assert that the list is not empty

PR #99, opened 2023-05-16
Increase startup probe to 3 minutes

peterrutg:
To increase the default from 2 to 3 minutes


SamTheisens:
> To increase the default from 2 to 3 minutes

@peterrutg 

> To increase the default from 2 to 3 minutes

[mpyl_config.example.yml](https://github.com/Vandebron/mpyl/pull/99/files#diff-f7ae7bfae1def422f340a5e021124625b524bec40576a9ea37370aaed6a6f683) is _just_ an example config. These specifics are no longer embedded in MP(y)L and you don't need a new release.
If you want to increase the default to 3 minutes you want to change that in the [monorepo](https://github.com/Vandebron/Vandebron/blob/master/mpyl_config.yml). Unless you think 2 minutes is a bad example default, you don't need to update anything in this repo.

PR #103, opened 2023-05-22
feature/TECH 300 tag deployments

koeves:
Monorepo counterpart: https://github.com/Vandebron/Vandebron/pull/10482

- Get changes in merge commit
- Add github action to create tag
- Jenkins logs are printed stdout to stdout with `--follow`

[TECH-300]: https://vandebron.atlassian.net/browse/TECH-300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TECH-300]: https://vandebron.atlassian.net/browse/TECH-300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mpyl-example.py
+++ b/mpyl-example.py
@@ -24,6 +24,7 @@ def main(log: Logger, args: argparse.Namespace):
         run_config=MpylRunConfig(config=config, run_properties=run_properties),
         parameters=MpylCliParameters(
             local=args.local,
+            tag=args.tag,
             pull_main=True,
             verbose=args.verbose,
             all=args.all
@@ -78,6 +79,7 @@ def main(log: Logger, args: argparse.Namespace):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description='Simple MPL pipeline')
     parser.add_argument('--local', '-l', help='a local developer run', default=False, action='store_true')
+    parser.add_argument('--tag', '-t', help='A tag run to deploy to test/acce/prod', default=False, action='store_true')
     parser.add_argument('--all', '-a', help='build and test everything, regardless of the changes that were made',
                         default=False, action='store_true')
     parser.add_argument('--dryrun', '-d', help="don't push or deploy images", default=False, action='store_true')
--- a/mpyl-example.py
+++ b/mpyl-example.py
@@ -79,7 +79,7 @@ def main(log: Logger, args: argparse.Namespace):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description='Simple MPL pipeline')
     parser.add_argument('--local', '-l', help='a local developer run', default=False, action='store_true')
-    parser.add_argument('--tag', '-t', help='A tag run to deploy to test/acce/prod', default=False, action='store_true')
+    parser.add_argument('--tag', '-t', help='The name of the tag to build', type=str)
     parser.add_argument('--all', '-a', help='build and test everything, regardless of the changes that were made',
                         default=False, action='store_true')
     parser.add_argument('--dryrun', '-d', help="don't push or deploy images", default=False, action='store_true')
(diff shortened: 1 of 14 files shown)


src/mpyl/utilities/repo/__init__.py
```diff
@@ -73,14 +73,21 @@ def latest_tag(self) -> str:
...
         in_branch = self.changes_in_branch()
         in_branch.append(Revision(len(in_branch), self.get_sha, self.changes_in_commit()))
         return in_branch
 
+    def changes_in_merge_commit(self) -> list[Revision]:
+        curr_rev = self._repo.rev_parse('HEAD')
+        parent_revs = reversed(list(curr_rev.iter_parents(n=1)))
+        return [Revision(count, str(rev),
+                         self._repo.git.diff_tree(no_commit_id=True, name_only=True, r=str(rev)).splitlines())
+                for count, rev in enumerate(parent_revs)]
```
SamTheisens, line 89: We assume that `HEAD` points to a tagged revision. Perhaps you could add a verification to make sure that you're indeed looking at the expected tag?
In that case you could also throw a clear error if we're not looking at a merge commit.

src/mpyl/utilities/jenkins/runner.py
```diff
@@ -159,4 +160,5 @@ def run(self):
 
         new_build_number = last_build_number + 1
 
-        self.follow_logs(job, new_build_number, last_build)
+        if self.follow:
+            self.follow_logs(job, new_build_number, last_build)
```
SamTheisens, line 164: Perhaps there's value in a `silent mode` too? Where you block (showing progress bar) until the build is done, but don't stream the logs.

Not for this PR, but just a thought.

src/mpyl/cli/commands/build/mpyl.py
```diff
@@ -29,6 +29,7 @@ class MpylRunConfig:
 @dataclass(frozen=True)
 class MpylCliParameters:
     local: bool
+    tag: bool = False
```
SamTheisens, line 32: Perhaps the tag could be an `Optional[str]` to validate (see later comment)

mpyl-example.py
```diff
@@ -78,6 +78,7 @@ def main(log: Logger, args: argparse.Namespace):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description='Simple MPL pipeline')
     parser.add_argument('--local', '-l', help='a local developer run', default=False, action='store_true')
+    parser.add_argument('--tag', '-t', help='A tag run to deploy to test/acce/prod', default=False, action='store_true')
```
SamTheisens, line 81: I think we probably need a target (test/acc/prod) parameter too. Not for this PR, but we will need it I think.

SamTheisens (commented):
LGTM in general. Some additional error handling would be good I think. I'm afraid this will be rather puzzling to people if things go wrong and the intent isn't clear.

SamTheisens (commented):
LGTM in general. Some additional error handling would be good I think. I'm afraid this will be rather puzzling to people if things go wrong and the intent isn't clear.

mpyl-example.py
koeves: It's handled by an environment variable in the run properties. Check also https://github.com/Vandebron/Vandebron/pull/10482

src/mpyl/utilities/jenkins/runner.py
koeves: I was doubting between the two approaches, if it should be silent or follow by default. But your version seems like a nice idea as well. Maybe let's hear what the users prefer as the default

src/mpyl/utilities/repo/__init__.py
koeves: My current thinking is that I can get the tag associated with the commit by doing `git describe --tags curr_rev`. This should match the $TAG_NAME in jenkins that we pass to mpyl by `--tag $TAG_NAME`. So indeed what you suggested, the tag will be a parameter not a flag

SamTheisens (approved):

PR #108, opened 2023-05-24
feature/TECH 428 any broken project yml should immediately fail the build

SamTheisens:
- Only log successful validation when `-v`
 - Remove team names from schema
 - Move maintainer list to config
 - Monorepo part https://github.com/Vandebron/Vandebron/pull/10494

### DIFF
--- a/src/mpyl/steps/steps.py
+++ b/src/mpyl/steps/steps.py
@@ -141,10 +141,21 @@ class Steps:
 
         return after_result
 
+    def _validate_project_against_config(self, project: Project) -> Optional[Output]:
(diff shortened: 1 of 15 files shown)


src/mpyl/steps/steps.py
```diff
@@ -141,11 +141,22 @@ def _execute_after_(self, main_result: Output, step: Step, project: Project, sta
 
         return after_result
 
+    def _validate_project_against_config(self, project: Project) -> Optional[Output]:
```
Jorg88, line 144: it's only validating maintainers right?

SamTheisens: Yes, for now we are

PR #109, opened 2023-05-24
Update ticket pattern defaults in schema and jira.py

MatthijsBon:
Sorry, I hadn't configured my git hooks yet 😕 

The Marketplace Growth team has a project with prefix "MG", which is two characters, meaning our ticket weren't recognized by mpyl. This fixes that.


src/mpyl/reporting/targets/jira.py
```diff
@@ -89,7 +89,7 @@ def from_config(config: dict):
         if not jira_config:
             raise KeyError('jira section needs to be defined in mpyl_config.yml')
         return JiraConfig(site=jira_config['site'], user_name=jira_config['userName'], password=jira_config['password'],
-                          ticket_pattern=re.compile(jira_config.get('ticketPattern', '[A-Za-z]{3,}-\\d+')),
+                          ticket_pattern=re.compile(jira_config.get('ticketPattern', '[A-Za-z]{2,}-\\d+')),
```
SamTheisens, line 92: This is just a fallback for when no pattern is configured. For the monorepo the pattern is configured here https://github.com/Vandebron/Vandebron/blob/c2e87c320916d74931639b1166f52289fd1a256f/mpyl_config.yml#L20.

SamTheisens (approved):
Looks good, but you'll need to update the monorepo too.

src/mpyl/reporting/targets/jira.py
MatthijsBon: Did: https://github.com/Vandebron/Vandebron/pull/10495 :)_

koeves, line 92: Can we add a unit test for this?

SamTheisens: The tests are there, just need to be linked up with the default config:
https://github.com/Vandebron/mpyl/blob/97a5d872414aae2d59c51db7e6a609c9acd8c5b9/tests/reporting/targets/test_jira.py#L45-L54

SamTheisens (approved):

PR #113, opened 2023-05-25
[TECH-434] Fix adding whitelists when none is defined for target

koeves:
There’s a whitelist target defined for TEST, but we’re deploying a PR therefore the configured whitelists `configured.get_value(self.target)` is None, which then throws an exception when the list is extended

Example run: https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline/view/change-requests/job/PR-10515/3/console

project.yml that triggered this: https://github.com/Vandebron/Vandebron/blob/c29489fd64c9a0e97345b4ce8c6183381c1ca784/energy-trading/curtailment/curtailment-volumes-api/deployment/project.yml#L90C1-L92


SamTheisens (approved):
👍

Jorg88 (approved):

PR #122, opened 2023-05-26
[TECH-128] Implementing a step to upload assets to S3 

benjaminlynch100:
branch: feature/TECH-128-cloudfront-kubernetes-deploy


src/mpyl/steps/deploy/cloudfront_kubernetes_deploy.py
```diff
@@ -0,0 +1,66 @@
...
+        container_path = f'{step_input.project.name}/{STATIC_FOLDER}'
+        tag = git_tag(step_input)
+
+        docker_copy(logger=self._logger, container_path=container_path, dst_path=TMP_FOLDER,
+                    image_name=image_name)
+        self.upload_to_s3(run_config=step_input.run_properties.config, bucket_name=bucket_name, src_path=TMP_FOLDER,
+                          bucket_root_path=tag)
+        delete_directory(logger=self._logger, dir_name=TMP_FOLDER)
+
+        return DeployKubernetes(self._logger).execute(step_input)
```
benjaminlynch100, line 39: Sam mentioned not the use the `after` property when calling the Kubernetes deploy step so I'm calling the execute directly here... not sure if this is the preferred approach, let me know!

src/mpyl/utilities/s3/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+                                    aws_secret_access_key=s3_config['secretAccessKey'])
+        self._bucket_name = bucket_name
+        self._root_path = root_path
+        self._logger = logger
+
+    def upload_directory(self, directory: str):
+        """
+        Uploads all files in a directory and its subdirectories to the S3 bucket
+
+        Note: boto3 does not provide a sync method, so we have to walk the directory and upload each file individually
```
benjaminlynch100, line 37: The lack of a sync is one of the main downsides of Amazons python SDK. There are other libraries that implement it, but it seems like a good idea to stick with the official SDK and it's not too big of a deal to get a similar result.

The only downside with this implementation is, with subsequent deployments to a PR environment, if an asset was deleted it will still remain in the bucket. I don't think this is a common enough occurrence to worry about and it only effects the test bucket (in Vandebron's case)

tests/steps/deploy/cloudfront/test_cloudfront.py
```diff
@@ -0,0 +1,12 @@
+from mpyl.steps.deploy.cloudfront_kubernetes_deploy import CloudFrontKubernetesDeploy
+from src.mpyl.steps.models import Input
+from tests.test_resources.test_data import get_cloudfront_project, RUN_PROPERTIES
+
+
+class TestCloudFront:
```
benjaminlynch100, line 6: it's tough to unit test the execute method here - it creates directories, copies files from docker containers, and uploads data to S3. I'll spend a little more time looking at this, but any recommendations?

koeves: I guess you can try mocking the calls and test the parameters that are passed around? Not sure if that's really feasible in python though. You can try moving as much as possible to utility functions and testing those?

koeves: And you can check as example the other deploy step tests

koeves: Ah I see that you mock the calls and moved everything to util methods, so maybe disregard my earlier comment :D I think it's fine like this

koeves (approved):
LGTM!
Don't forget to merge https://github.com/Vandebron/mpyl_config/pull/1 as well after this, I moved your S3 config change here

Jorg88 (commented):
Nice job in general, very thorough on the documentation as well!

koeves (changes requested):
Checked the unit tests again, one suggestion about the `test_filesystem`:

There you are using the boolean return values, but the tests themselves are mostly just testing if the mock works correctly..
Maybe you could:
1. remove those test
2. Make sure all the boolean return values are propagated up to the deploy step, docker, s3 client etc
3. Create more thorough unit tests in those modules where external calls are mocked and test if the flow is correct

Like `docker_copy` should also return False if `create_directory` does, and in turn execute should fail as well

So then we can add two test, one in docker and one the deploy step to check these?

src/mpyl/steps/deploy/cloudfront_kubernetes_deploy.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def upload_to_s3(self, run_config: Dict, bucket_name: str, src_path: str, bucket_root_path: str):
```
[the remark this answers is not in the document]
benjaminlynch100: will update!

SamTheisens:
Looks great in general! I like the fact that you've wrapped the S3 client. First use of mocking for this project 👍. Although it's generally best to keep mocking to a mininimum, I think there are currently some untested bits that we could test with the help of patch.


For temp folders: 
https://github.com/Vandebron/mpyl/pull/122/files#diff-d15968178de9592b92d529099660ebe7dbf683c7f1d25205bec0ac16d42e0bafR33-R37
you can use https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory using a `with` (try with resoruces), so the temp files are guaranteed to be cleaned up

I think this a good way to chain steps :thumbs: https://github.com/Vandebron/mpyl/pull/122/files#diff-d15968178de9592b92d529099660ebe7dbf683c7f1d25205bec0ac16d42e0bafR39

For https://github.com/Vandebron/mpyl/pull/122/files#diff-e8b4ceffe17bb0b13add37010de17c46af0323088877b5afccd5b6d750cf6142R83-R85 you can use https://github.com/Vandebron/mpyl/blob/a541debbc05becb2f22c6058e7c8bb53f1c09159/src/mpyl/steps/models.py#L29-L30


The logic at https://github.com/Vandebron/mpyl/pull/122/files#diff-e8b4ceffe17bb0b13add37010de17c46af0323088877b5afccd5b6d750cf6142R92-R104 is very similar to https://github.com/Vandebron/mpyl/blob/9b0151d9d22cf6f2c689793fa785725c06c8e692/src/mpyl/steps/test/dockertest.py#L79-L88, maybe you can find a way to reuse. You can probably also use container id as a unique identifier (like here: https://github.com/Vandebron/mpyl/blob/9b0151d9d22cf6f2c689793fa785725c06c8e692/src/mpyl/steps/test/dockertest.py#L74)

Do we need wrappers around such fundamental operations? https://github.com/Vandebron/mpyl/pull/122/files#diff-2142665a90e5a80a6625be7c158fac3ab9473a6b7c9fc9bc77a820c98497bef1R8 Rmtree for example already has a parameter to ignore errors: https://github.com/Vandebron/mpyl/pull/122/files#diff-2142665a90e5a80a6625be7c158fac3ab9473a6b7c9fc9bc77a820c98497bef1R28


WRT https://github.com/Vandebron/mpyl/pull/122/files#diff-96f40a0494c735a518dcd08ad1cdb2b5faa5d056475d286c333b74324d3a3d13R13. The pattern we've been following so far is to use pure data classes for config. See here for example: https://github.com/Vandebron/mpyl/blob/ae36365ef97697b2942aee6811f49efaf35701cc/src/mpyl/utilities/github/__init__.py. The preferred way is to use a factory method like https://github.com/Vandebron/mpyl/blob/ae36365ef97697b2942aee6811f49efaf35701cc/src/mpyl/utilities/github/__init__.py#L34.
I don't think we need to replicate the requiredness checks that are already in the schema, but more complicated validations can still be done in code and throw a `KeyError` when appropriate.

In a couple of locations i've noticed you capture exceptions and return a boolean instead. Why not let the exceptions bubble up?

src/mpyl/steps/deploy/cloudfront_kubernetes_deploy.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+        Creates an S3 client and uploads the static assets stored in the temp folder
+
+        :param run_config: the config containing S3 info
+        :param bucket_name: the name of the bucket
```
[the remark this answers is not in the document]
benjaminlynch100: there was no particular reason I did it that way - the bucket name is now retrieved in the new S3Config class

src/mpyl/steps/steps.py
```diff
@@ -87,11 +88,12 @@ def __init__(self, logger: Logger, properties: RunProperties) -> None:
                 TestDocker(logger)
             },
             Stage.DEPLOY: {
+                CloudFrontKubernetesDeploy(logger),
                 DeployEcho(logger),
                 DeployKubernetes(logger),
                 DeployKubernetesJob(logger),
                 DeployKubernetesSparkJob(logger),
-                EphemeralDockerDeploy(logger)
+                EphemeralDockerDeploy(logger),
```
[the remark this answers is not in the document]
benjaminlynch100: fixed

src/mpyl/utilities/filesystem/__init__.py
```diff
@@ -0,0 +1,41 @@
+"""Filesystem related utility methods"""
+
+import os
+import shutil
+from logging import Logger
+
+
+def create_directory(logger: Logger, dir_name: str, overwrite=True) -> bool:
```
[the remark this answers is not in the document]
benjaminlynch100: I've deleted this file 😄

src/mpyl/utilities/docker/__init__.py
```diff
@@ -75,15 +76,34 @@ def stream_docker_logging(logger: Logger, generator: Union[Iterator[str], Iterat
...
+
+    :param logger: the logger
+    :param container_path: the path of the directory in the container to copy
+    :param dst_path: the path to copy the container content to
+    :param image_name: the name of the docker image which a container is created from
+    """
+    create_directory(logger=logger, dir_name=dst_path)
+    container = docker.create(image_name, name='container')
+    docker.copy((container.name, container_path), dst_path)
+    docker.remove(container)
```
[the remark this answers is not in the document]
benjaminlynch100: I initially removed the container because I had created the container from the image within this function. I've completely changed this method now following @SamTheisens comment - it's using the same logic that was in the dockerTest file

src/mpyl/steps/deploy/cloudfront_kubernetes_deploy.py
```diff
@@ -0,0 +1,66 @@
...
+        self._logger.info('Upload complete')
+
+    @staticmethod
+    def get_bucket_name(step_input: Input) -> str:
+        """
+        Retrieves the S3 bucket name from the step input
+
+        """
+        s3_config = step_input.project.s3_bucket
+        if s3_config is None or s3_config.bucket is None:
```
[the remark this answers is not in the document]
benjaminlynch100: my understanding, and the reason this is different to the init method, is that the `s3_bucket` is an optional property on the Project class, as opposed to being in a dictionary

src/mpyl/steps/deploy/cloudfront_kubernetes_deploy.py
```diff
@@ -0,0 +1,54 @@
...
+from .. import Step, Meta
+from ...project import Stage
+from ..models import Input, Output, ArtifactType
+from ...utilities.docker import docker_image_tag, docker_copy
+from ...utilities.s3 import S3Client, S3ClientConfig
+
+STATIC_FOLDER = 'static'
+
+
+class CloudFrontKubernetesDeploy(Step):
```
ewaldschrader, line 16: Why still the name CloudFrontKubernetesDeploy? Why nog something like DeployS3?

benjaminlynch100: I originally thought that would be a more appropriate name but since the step also calls the Kubernetes Deploy step, and the s3 key structure is specific to our cloudfront setup, it seems to be the best name

benjaminlynch100:
There was an [update](https://github.com/Vandebron/mpyl/pull/139) that allows the mpyl config branch to specified in the test pipeline - after rebasing on main I still wasn't able to specify the branch with this PR so I created a new branch & PR

I've addressed all the previous comments within the new PR: https://github.com/Vandebron/mpyl/pull/141

PR #123, opened 2023-05-27
fix/TECH 438 usability issues with m py l

SamTheisens:
- [x] make version check non-blocking, to prevent mpyl CLI from becoming sluggish with slow internet connection
- [x] `--follow` by default, as it it's more instructive for first time use
- [x] `--silent` mode, in which we block until the Jenkins build is finished but filter out the logs  
- [x] add logging and documentation for perhaps less than intuitive "copy test results form image" step 
- [x] fix link to github from terminal 
- [x] only switch assignment if not yet
- [x] rename `.mpl` folder to `.mpyl`
- [x] introduce possibility to filter documentation changes from invalidation logic: https://vandebron.slack.com/archives/CGJDDBETU/p1685096548855339


src/mpyl/steps/test/dockertest.py
```diff
@@ -1,9 +1,23 @@
...
+
+
+## 🧪 Testing inside a container
+
+When unit tests are run within a docker container the test results need to be written to a folder inside it.
+This means that the test step _within the docker container_ should not return a system error.
+Otherwise, building of the container would stop and the test results would not be committed to a layer.
+
+The test results need to be writted  written to a folder named `$WORKDIR/target/test-reports/` for
+`TestDocker.extract_test_results` to find and extract them.
```
koeves, line 11: Typo here
```suggestion
The test results need to be written to a folder named `$WORKDIR/target/test-reports/` for
`TestDocker.extract_test_results` to find and extract them.
```

src/mpyl/utilities/jenkins/runner.py
```diff
@@ -89,52 +95,68 @@ def await_parameter_build(self, build_job: Job):
...
+
+        build_to_follow.block_until_complete()
+        finished_build = job.get_last_build()
+        self.status.console.log(
+            f'[link={finished_build.get_build_url()}][i]Build[/link] for {self.pipeline.human_readable()} '
+            f'ended with outcome {self.to_icon(finished_build)}', markup=True)
+        self.status.console.log()
+
+        os.system('afplay /System/Library/Sounds/' + ('Glass.aiff' if finished_build.is_good() else 'Sosumi.aiff'))
+        sys.exit()
```
koeves, line 120: So we expect mpyl to be run on a mac?

koeves (approved):
Nice improvements!

src/mpyl/utilities/jenkins/runner.py
SamTheisens: ah, yeah, good one. We should add a condition here.

PR #138, opened 2023-06-14
[TECH-394] Add multiple links to the cypress dashboard if multiple su…

Jorg88:
…ites have run

branch: feature/TECH-394-allow-multiple-cypress-result-urls
----
📕 [TECH-394](https://vandebron.atlassian.net/browse/TECH-394) Integrate cypress suites into mpyl jorgpost@vandebron.nl 
Make sure that certain cypress suites get triggered for certain projects when specified in the project.yml.

🏗️ Build [12](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-138/12/display/redirect) ✅ Successful, started by _Jorg Post_  
🚀  *[cloudfront-service](https://cloudfront-service-138.test.nl/)*, *enrichChargeSessionsJob*, *job*, *kong-sync*, *[nodeservice](https://nodeservice-138.test.nl/)*, *[sbtservice](https://sbtservice-138.test.nl/)*  

[TECH-394]: https://vandebron.atlassian.net/browse/TECH-394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/reporting/formatting/test_resources/markdown_run.md
+++ b/tests/reporting/formatting/test_resources/markdown_run.md
@@ -6,5 +6,5 @@ Build failed
 ```
 🏗️  _dockertest_, _test_  
 📋  _test_  
-🧪 51 ❌ 1 💔 0 🙈 0 [link](http://localhost/tests) 
+🧪 51 ❌ 1 💔 0 🙈 0 [Docker Test](http://localhost/tests)  
 🚀  _test_  
--- a/src/mpyl/reporting/formatting/markdown.py
+++ b/src/mpyl/reporting/formatting/markdown.py
@@ -2,14 +2,17 @@
 Markdown run result formatters
 """
 import operator
+import itertools
 
 from junitparser import TestSuite
 
 from ...project import Stage, Project
 from ...steps import Output, ArtifactType
+from ...steps.models import Artifact
 from ...steps.run import RunResult
-from ...steps.steps import StepResult, collect_test_results, collect_test_artifacts
-from ...utilities.junit import TestRunSummary, sum_suites, TEST_RESULTS_URL_KEY
+from ...steps.steps import StepResult
+from ...utilities.junit import TestRunSummary, sum_suites, TEST_RESULTS_URL_KEY, TEST_RESULTS_URL_NAME_KEY, \
+    to_test_suites
 
 
 def summary_to_markdown(summary: TestRunSummary):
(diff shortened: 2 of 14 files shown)


koeves (approved):

SamTheisens:
Would it be an idea to generically write a list of `test_url` to the output? I think that might be cleaner than making a a special case for cypress tests in a generic formatter.

Jorg88:
> Would it be an idea to generically write a list of `test_url` to the output? I think that might be cleaner than making a a special case for cypress tests in a generic formatter.

You mean taking the current `tests_url` out of the `RunContext` and into the `RunResult`, turning it into a list[str] and only setting it when tests are run?

tests/reporting/formatting/test_resources/markdown_run.md
```diff
@@ -6,5 +6,5 @@ Build failed
 ```
 🏗️  _dockertest_, _test_  
 📋  _test_  
-🧪 51 ❌ 1 💔 0 🙈 0 [link](http://localhost/tests) 
+🧪 51 ❌ 1 💔 0 🙈 0 [Docker Test](http://localhost/tests)
```
Jorg88, line 9: The extra spaces are on purpose, it fixes an issue where locally the lines after the results weren't on a newline. The other lines also have 2 spaces

SamTheisens (approved):
Indeed, that's what I meant 👍. Perhaps at some point we should create a dataclass which we can serialize instead of the somehwat brittle and untransparent use of keys in a dictionary. But I guess that can wait.

Jorg88:
> Indeed, that's what I meant 👍. Perhaps at some point we should create a dataclass which we can serialize instead of the somehwat brittle and untransparent use of keys in a dictionary. But I guess that can wait.

Yeah the `spec` of `Artifact` could/should be a dataclass indeed.

However, I found a new issue with the current solution 😅 for the unit tests it will now link the same url (since they are always on the same page) for every project. I'm thinking about a solution where I merge them when the url's are the same, but it won't be as 'clean' as the current way.

SamTheisens:
@Jorg88 set([]) takes care of duplicates and is not ugly I think?

Jorg88:
> @Jorg88 set([]) takes care of duplicates and is not ugly I think?

Yeah but the dict's are not entirely similar, only the url's, so I had to do something else. Let me know what you think about the current solution!

koeves (approved):

PR #145, opened 2023-06-23
Add python setup to docs workflow

koeves:
----
 # ⚠️ Could not find ticket corresponding to `fix/docs_generation. Does your branch name follow the correct pattern?

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,7 +4,7 @@ name: docs
 on:
   push:
     branches:
-      - main
+    - main
 
 # security: restrict permissions for CI jobs.
 permissions:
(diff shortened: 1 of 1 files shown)


Jorg88 (approved):

.github/workflows/docs.yml
```diff
@@ -15,25 +15,26 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - name: Install dependencies
-        uses: VaultVulp/action-pipenv@v2.0.1
-        with:
-          command: install -d
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
```
koeves, line 19: This line @Jorg88

PR #149, opened 2023-06-26
Prepare for customizable stages

SamTheisens:
Replacing the Stage enum with a dynamic set of stages that can be configured via `mpyl_config` requires significant refactorings.
In order to limit the scope of the work, this PR concerns itself with preparational work and small bugfixes only. In particular:

 - [x] Rewrite the build github action so the different CI stages may be run in parallel
 - [x] Discover step executors via the plugin mechanism instead of requiring them to be hardcoded in order to make it easier to locally define custom steps
 - [x] Remove .egg and .tar.gz distribution formats because pypi no longer accepts them
 - [x] Create end-to-end build test, to have more confidence about the execution mechanism at unit test time
 - [x] Prevent exceptions thrown in the scope of a Repostirory try with resources from being swallowed
 - [x] Attempt to fix local builds via the `mpyl` CLI in the root.

Note: `.env` in the monorepo should not contain `MPYL_CONFIG_PATH=deployment/pipeline/mpyl/config.yml` in case it still does. Perhaps you could make sure that `mpyl build status` and `mpyl health` still work on the monorepo.

----
### 📕 [TECH-293](https://vandebron.atlassian.net/browse/TECH-293) As an MPyL user I want to define my own stage(s) ewaldschrader@vandebron.nl 
At the moment stages are a hardcoded concept in MPyL [steps.py#L60-L85](https://github.com/Vandebron/mpyl/blob/37c7b0c9a7e9eb54dfb05d8822c4e360469855b7/src/mpyl/steps/steps.py#L60-L85) 

While there is potential order dependency between build steps, a fixed order is too limiting. For example, a *security scan_ or _integration test* stage could be run in parallel after the build stage. Defining the execution flow is more at home in the (python) definition of the flow than within the library.

An important aspect to consider is the invalidation logic, which is stage specific (at the moment) [discovery.py#L41-L43](https://github.com/Vandebron/mpyl/blob/e0a8fabcf5c08492f0382bae4df8d5502b45d155/src/mpyl/stages/discovery.py#L41-L43). Perhaps this could be converted to a strategy.

🏗️ Build [13](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-149/13/display/redirect) ✅ Successful, started by _Daniel Koves_  
🚀 *[cloudfront-service](https://cloudfront-service-149.test.nl/)*, *enrichChargeSessionsJob*, *job*, *kong-sync*, *[nodeservice](https://nodeservice-149.test.nl/)*, *[sbtservice](https://sbtservice-149.test.nl/)*  

[TECH-293]: https://vandebron.atlassian.net/browse/TECH-293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/mpyl/cli/commands/build/mpyl.py
+++ b/src/mpyl/cli/commands/build/mpyl.py
@@ -70,7 +70,6 @@ def run_mpyl(mpyl_run_parameters: MpylRunParameters, reporter: Optional[Reporter
     console_properties = mpyl_run_parameters.run_config.run_properties.console
     console = Console(markup=False, width=None if params.local else console_properties.width, no_color=False,
                       log_path=False, color_system='256')
-    print(f"Logging properties: {console_properties}, console: {console}")
     logging.raiseExceptions = False
     logging.basicConfig(
         level="DEBUG" if params.verbose else console_properties.log_level, format=FORMAT, datefmt="[%X]",
--- a/src/mpyl/cli/commands/build/mpyl.py
+++ b/src/mpyl/cli/commands/build/mpyl.py
@@ -126,18 +126,22 @@ def find_build_set(repo: Repository, changes_in_branch: list[Revision], build_al
 
 
 def run_build(accumulator: RunResult, executor: Steps, reporter: Optional[Reporter] = None, dry_run: bool = True):
-    for stage, projects in accumulator.run_plan.items():
-        for proj in projects:
-            result = executor.execute(stage, proj, dry_run)
-            accumulator.append(result)
-            if reporter:
-                reporter.send_report(accumulator)
+    try:
+        for stage, projects in accumulator.run_plan.items():
+            for proj in projects:
+                result = executor.execute(stage, proj, dry_run)
+                accumulator.append(result)
+                if reporter:
+                    reporter.send_report(accumulator)
 
-            if not result.output.success and stage == Stage.DEPLOY:
-                logging.warning(f'Deployment failed for {proj.name}')
+                if not result.output.success and stage == Stage.DEPLOY:
+                    logging.warning(f'Deployment failed for {proj.name}')
+                    return accumulator
+
+            if accumulator.failed_result:
+                logging.warning(f'One of the builds failed at {stage}')
                 return accumulator
-
-        if accumulator.failed_result:
-            logging.warning(f'One of the builds failed at {stage}')
-            return accumulator
-    return accumulator
+        return accumulator
+    except ExecutionException as exc:
+        accumulator.exception = exc
+        return accumulator
--- a/src/mpyl/reporting/targets/slack.py
+++ b/src/mpyl/reporting/targets/slack.py
@@ -107,13 +107,15 @@ class SlackReporter(Reporter):
 
     def send_report(self, results: RunResult, text: Optional[str] = None) -> ReportOutcome:
         try:
-            user_info = self.__get_user_info(results.run_properties.details.user_email)
+            email = results.run_properties.details.user_email
+            user_info = self.__get_user_info(email)
 
             if not self._channel and user_info.initiator:
                 self._channel = self.__open_conversation_with_user(user_info.initiator)
 
             if not self._channel:
-                raise ValueError('Channel not explicitly set and initiator could not be determined')
+                return SlackOutcome(success=False, exception=ValueError(
+                    f'Channel not explicitly set and initiator for {email} could not be determined'))
 
             body = to_slack_markdown(text if text else run_result_to_markdown(results))
             blocks = self.__compose_blocks(results, body, user_info)
--- a/src/mpyl/steps/collection.py
+++ b/src/mpyl/steps/collection.py
@@ -13,8 +13,7 @@ class StepsCollection:
 
     def __init__(self, logger: Logger, base_path: Optional[str] = None) -> None:
         self._step_executors = set()
-        self._base_path = base_path
-        self.__load_steps_in_module('.')
+        self.__load_steps_in_module('.', base_path)
 
         for plugin in IPluginRegistry.plugins:
             step_instance: Step = plugin(logger)
@@ -22,8 +21,9 @@ class StepsCollection:
             logger.debug(f"{meta.name} for stage {meta.stage} registered. Description: {meta.description}")
             self._step_executors.add(step_instance)
 
-    def __load_steps_in_module(self, module_root: str):
-        module = importlib.import_module(module_root, f'{self._base_path + "." if self._base_path else ""}mpyl.steps')
+    @staticmethod
+    def __load_steps_in_module(module_root: str, base_path: Optional[str] = None):
+        module = importlib.import_module(module_root, f'{base_path + "." if base_path else ""}mpyl.steps')
         for _, modname, _ in pkgutil.walk_packages(path=module.__path__, prefix=module.__name__ + '.',
                                                    onerror=lambda x: None):
             importlib.import_module(modname)
--- a/src/mpyl/steps/collection.py
+++ b/src/mpyl/steps/collection.py
@@ -1,0 +1,44 @@
+"""A collection of all available step executors."""
+from logging import Logger
+from typing import Optional
+
+from . import Step
+from .build.dockerbuild import BuildDocker
+from .build.echo import BuildEcho
+from .build.sbt import BuildSbt
+from .deploy.cloudfront_kubernetes_deploy import CloudFrontKubernetesDeploy
+from .deploy.echo import DeployEcho
+from .deploy.ephemeral_docker_deploy import EphemeralDockerDeploy
+from .deploy.kubernetes import DeployKubernetes
+from .deploy.kubernetes_job import DeployKubernetesJob
+from .deploy.kubernetes_spark_job import DeployKubernetesSparkJob
+from .postdeploy.cypress_test import CypressTest
+from .test.dockertest import TestDocker
+from .test.echo import TestEcho
+from .test.sbt import TestSbt
+from ..project import Stage
+
+
+class StepsCollection:
+    _step_executors: set[Step]
+
+    def __init__(self, logger: Logger) -> None:
+        self._step_executors = {
+            BuildEcho(logger),
+            BuildSbt(logger),
+            BuildDocker(logger),
+            TestEcho(logger),
+            TestSbt(logger),
+            TestDocker(logger),
+            CloudFrontKubernetesDeploy(logger),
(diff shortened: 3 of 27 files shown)


Pipfile
```diff
@@ -46,7 +46,7 @@ python_version = "3.9"
 [scripts]
 cli = "python src/mpyl/__main__.py"
 cli-ext = "bash -c \"python $(dirname \\$PIPENV_PIPFILE)/src/mpyl/__main__.py $(echo $0) $(echo $@)\""
-build = "python setup.py install bdist bdist_wheel"
+build = "python setup.py bdist_wheel"
```
Jorg88, line 49: Is this change based on the e-mail we got from pypi?

src/mpyl/cli/commands/build/mpyl.py
```diff
@@ -126,18 +125,22 @@ def find_build_set(repo: Repository, changes_in_branch: list[Revision], build_al
...
-                logging.warning(f'Deployment failed for {proj.name}')
+    try:
+        for stage, projects in accumulator.run_plan.items():
+            for proj in projects:
+                result = executor.execute(stage, proj, dry_run)
+                accumulator.append(result)
+                if reporter:
+                    reporter.send_report(accumulator)
+
+                if not result.output.success and stage == Stage.DEPLOY:
```
Jorg88, line 136: You probably noticed already, but we made sure other stages now only fail after finishing all the projects. Might be something we have to make configurable as well

src/mpyl/steps/collection.py
```diff
@@ -0,0 +1,33 @@
...
+from . import Step, IPluginRegistry
+from ..project import Stage
+
+
+class StepsCollection:
+    _step_executors: set[Step]
+
+    def __init__(self, logger: Logger, base_path: Optional[str] = None) -> None:
+        self._step_executors = set()
+        self._base_path = base_path
```
Jorg88, line 16: Just curious: is there a reason `_step_executors` gets initiated above but `_base_path` not?

src/mpyl/steps/models.py
```diff
@@ -107,7 +107,7 @@ def from_configuration(run_properties: Dict, config: Dict):
 
         return RunProperties(
             details=RunContext.from_configuration(build['run']),
-            target=Target(build['parameters']['deploy_target']),
+            target=Target(build['parameters']['deploy_target'] or Target.PULL_REQUEST.value),
```
Jorg88, line 110: This feels a bit hacky but it's probably just my limited python knowledge. It's creating a new `Target` from the value of an already existing one. Plus the `or` only covers the scenario in which 'deploy_target' is `None` right? But we made it optional above, so it could also not exist at all and give a KeyError anyway.

src/mpyl/steps/steps.py
```diff
@@ -52,51 +40,20 @@ class StepResult:
...
-    _step_executors: dict[Stage, set[Step]]
     _logger: Logger
     _properties: RunProperties
 
-    def __init__(self, logger: Logger, properties: RunProperties) -> None:
+    def __init__(self, logger: Logger, properties: RunProperties,
+                 steps_collection: Optional[StepsCollection] = None) -> None:
+        self._logger = logger
+        self._properties = properties
+        self._steps_collection = steps_collection or StepsCollection(logger)
```
Jorg88, line 50: 🚀

tests/cli/commands/test_build.py
```diff
@@ -0,0 +1,70 @@
...
+        assert result.status_line == '🦥 Nothing to do'
+
+    def test_run_build_with_plan_should_execute_successfully(self):
+        run_properties = RUN_PROPERTIES
+
+        projects = [get_minimal_project()]
+        run_plan = {
+            Stage.BUILD: projects,
+            Stage.TEST: projects,
+            Stage.DEPLOY: projects
```
Jorg88, line 42: did you leave out post_deploy on purpose?

Jorg88 (commented):
Nice work Sam!

If you're going to continue working on MPyL every now and then, should we occasionally have some sort of sync where we can catch up and you can share some of your ideas & feedback? I'll leave it up to you since your committed time is completely voluntary!

SamTheisens:
- https://github.com/Vandebron/mpyl/pull/149#discussion_r1243280643
   Yes, because the minimal test project doesn't have a postdeploy stage. We _could_ create a dummy post deploy step, but I'm not sure what we would gain from that.
 - https://github.com/Vandebron/mpyl/pull/149#discussion_r1243235670
   Yes, I had noticed that. I find it a bit counterintuitive as a default to be honest. I would expect the default build to "fail fast".
   I also think we should try to avoid `stage == Stage.WHATEVER` conditionals as much as possible, as it results in implicit behavior and reduced the flexibility/abstraction of the stage concept. I believe these conditions belong in the logic of the runner, so external to MPyL. Making the stages configurable is a step in a direction away from the embedded "default" runner. WDYT?
 - https://github.com/Vandebron/mpyl/pull/149#discussion_r1243234237
   The `install` parameter creates the `.egg`, and `bdist` creates a `tar.gz`, which https://test.pypi.org has started to reject, complaning about naming. The latter is a problem we could try to solve, but I don't think we need any other distribution format than `.wheel`

The other comments are fixed (hopefully).

Note: the jira ticket update seems to have messed up the github comment. Any reason the ticket title is no longer a header/bold? I did find it a bit more readable before tbh.

Regarding having a sync some time. I'd be happy to, although I can't really commit to any particular interval or duration. Maybe we can talk on Whatsapp?

koeves:
>   I also think we should try to avoid `stage == Stage.WHATEVER` conditionals as much as possible, as it results in implicit behavior and reduced the flexibility/abstraction of the stage concept. I believe these conditions belong in the logic of the runner, so external to MPyL. Making the stages configurable is a step in a direction away from the embedded "default" runner. WDYT?

I agree, with made the quickfix choice until we refactor the stages (now).

> Note: the jira ticket update seems to have messed up the github comment. Any reason the ticket title is no longer a header/bold? I did find it a bit more readable before tbh.

I noticed that as well lately. I changed the ticket title heading level in a previous card, maybe that messed this up.

> Regarding having a sync some time. I'd be happy to, although I can't really commit to any particular interval or duration. Maybe we can talk on Whatsapp?

I'm fine with using whatsapp

src/mpyl/reporting/targets/slack.py
```diff
@@ -107,13 +107,15 @@ def __init__(self,
...
-            user_info = self.__get_user_info(results.run_properties.details.user_email)
+            email = results.run_properties.details.user_email
+            user_info = self.__get_user_info(email)
 
             if not self._channel and user_info.initiator:
                 self._channel = self.__open_conversation_with_user(user_info.initiator)
 
             if not self._channel:
-                raise ValueError('Channel not explicitly set and initiator could not be determined')
+                return SlackOutcome(success=False, exception=ValueError(
```
Jorg88, line 117: Seems the build fails on this, which means pr's that are build by you cannot be merged. It should warn but not fail the build, right?

koeves:
@SamTheisens going to merge this and the black formatter PR today, I tested them with my open PRs in the monorepo

SamTheisens:
@koeves cool 👍

SamTheisens:
@koeves cool 👍

koeves:
Btw did this PR require any changes in the mpyl repo that is not committed? Currently if I do pipenv run run it fails with:
```
ModuleNotFoundError: No module named 'mpyl'
           WARNING  mpyl  Unexpected exception: No module named 'mpyl'                                                                              mpyl-example.py:148
                    Traceback (most recent call last):                                                                                                                 
                      File "/Users/danielkoves/vdb/mpyl/mpyl-example.py", line 146, in <module>                                                                        
                        main(mpl_logger, parsed_args)                                                                                                                  
                      File "/Users/danielkoves/vdb/mpyl/mpyl-example.py", line 91, in main                                                                             
                        run_result = run_mpyl(params, slack_personal)                                                                                                  
                      File "/Users/danielkoves/vdb/mpyl/src/mpyl/cli/commands/build/mpyl.py", line 111, in run_mpyl                                                    
                        raise exc                                                                                                                                      
                      File "/Users/danielkoves/vdb/mpyl/src/mpyl/cli/commands/build/mpyl.py", line 106, in run_mpyl                                                    
                        return run_result                                                                                                                              
                      File "/Users/danielkoves/vdb/mpyl/src/mpyl/utilities/repo/__init__.py", line 79, in __exit__                                                     
                        raise exc_val                                                                                                                                  
                      File "/Users/danielkoves/vdb/mpyl/src/mpyl/cli/commands/build/mpyl.py", line 95, in run_mpyl                                                     
                        steps = Steps(logger=logger, properties=mpyl_run_parameters.run_config.run_properties)                                                         
                      File "/Users/danielkoves/vdb/mpyl/src/mpyl/steps/steps.py", line 55, in __init__                                                                 
                        self._steps_collection = steps_collection or StepsCollection(logger)                                                                           
                      File "/Users/danielkoves/vdb/mpyl/src/mpyl/steps/collection.py", line 16, in __init__                                                            
                        self.__load_steps_in_module(".", base_path)                                                                                                    
                      File "/Users/danielkoves/vdb/mpyl/src/mpyl/steps/collection.py", line 28, in __load_steps_in_module                                              
                        module = importlib.import_module(                                                                                                              
                      File "/opt/homebrew/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/importlib/__init__.py",                      
                    line 127, in import_module                                                                                                                         
                        return _bootstrap._gcd_import(name[level:], package, level)                                                                                    
                      File "<frozen importlib._bootstrap>", line 1030, in _gcd_import                                                                                  
                      File "<frozen importlib._bootstrap>", line 1007, in _find_and_load                                                                               
                      File "<frozen importlib._bootstrap>", line 972, in _find_and_load_unlocked                                                                       
                      File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed                                                                     
                      File "<frozen importlib._bootstrap>", line 1030, in _gcd_import                                                                                  
                      File "<frozen importlib._bootstrap>", line 1007, in _find_and_load                                                                               
                      File "<frozen importlib._bootstrap>", line 984, in _find_and_load_unlocked                                                                       
                    ModuleNotFoundError: No module named 'mpyl'
```

this happens on `mpyl.steps.collection.StepsCollection.__load_steps_in_module`

PR #155, opened 2023-06-30
Fix getting changes for single commit branches

koeves:
- Fix getting changes for branches with one commit
- Put Acceptance back as valid target

----
 # ⚠️ Could not find ticket corresponding to `feature/fix_getting_changes. Does your branch name follow the correct pattern?

### DIFF
--- a/src/mpyl/schema/run_properties.schema.yml
+++ b/src/mpyl/schema/run_properties.schema.yml
@@ -55,7 +55,7 @@ properties:
           deploy_target:
             description: The deploy target.
             type: ["string", "null"]
-            enum: [ null,  'PullRequest', 'PullRequestBase', 'Production' ]
+            enum: [ null,  'PullRequest', 'PullRequestBase', 'Acceptance', 'Production' ]
       console:
         type: object
         additionalProperties: false
--- a/src/mpyl/utilities/repo/__init__.py
+++ b/src/mpyl/utilities/repo/__init__.py
@@ -141,7 +141,7 @@ class Repository:
         if not revisions:
             return []
 
-        first_revision = revisions[0].hexsha if len(revisions) != 1 else "master"
+        first_revision = revisions[0].hexsha if len(revisions) != 1 else self._config.main_branch
 
         files_touched_in_branch = set(
             self._repo.git.diff(
--- a/src/mpyl/utilities/repo/__init__.py
+++ b/src/mpyl/utilities/repo/__init__.py
@@ -141,7 +141,9 @@ class Repository:
         if not revisions:
             return []
 
-        first_revision = revisions[0].hexsha if len(revisions) != 1 else self._config.main_branch
+        first_revision = (
+            revisions[0].hexsha if len(revisions) != 1 else self._config.main_branch
+        )
 
         files_touched_in_branch = set(
             self._repo.git.diff(
--- a/src/mpyl/utilities/repo/__init__.py
+++ b/src/mpyl/utilities/repo/__init__.py
@@ -138,6 +138,11 @@ class Repository:
         revisions = list(
             reversed(list(self._repo.iter_commits(f"{self._config.main_branch}..HEAD")))
         )
+
+        logging.debug(
+            f"Found {len(revisions)} revisions in branch: {[r.hexsha for r in revisions]}"
+        )
+
         if not revisions:
             return []
 
--- a/src/mpyl/utilities/repo/__init__.py
+++ b/src/mpyl/utilities/repo/__init__.py
@@ -141,9 +141,11 @@ class Repository:
         if not revisions:
             return []
 
+        first_revision = revisions[0].hexsha if len(revisions) != 1 else "master"
+
         files_touched_in_branch = set(
             self._repo.git.diff(
-                f"{revisions[0].hexsha}..{revisions[-1].hexsha}", name_only=True
+                f"{first_revision}..{revisions[-1].hexsha}", name_only=True
             ).splitlines()
         )
         return [


koeves:
@SamTheisens do you know the reason for that change some time ago? Currently it breaks if you only have one commit in your branch, because then we diff the first commit with itself (the hex[0]...hex[-1]) part:
```
DEBUG    git.cmd  Popen(['git', 'diff', '--name-only', 'cdcf695d68066c949388cf95331b0663ff8227c2..cdcf695d68066c949388cf95331b0663ff8227c2'], 
```
https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline/job/PR-10840/6/consoleFull (not sure if you have access to jenkins)

SamTheisens:
@koeves that's because of this: https://github.com/Vandebron/mpyl/commit/9346192ebfe3f2b6b4f5e85e1c952b0bb04b4ce4. Even if main has never been checked out, we are still able to find out what happened on the current branch.
A fix for the "single (or no) commit on branch is in the PR here https://github.com/Vandebron/mpyl/pull/149/commits/1f1a880c0da4acabb52c877131445b78fc9bed44, besides a (very important) fix for making sure problems related to the repo aren't swallowed silently https://github.com/Vandebron/mpyl/pull/149/commits/ea351c5d7715d89bfaafd33a2acf49d6ea28a224

koeves:
> @koeves that's because of this: [9346192](https://github.com/Vandebron/mpyl/commit/9346192ebfe3f2b6b4f5e85e1c952b0bb04b4ce4). Even if main has never been checked out, we are still able to find out what happened on the current branch. A fix for the _single (or no) commit on branch_ bug is in the PR here [1f1a880](https://github.com/Vandebron/mpyl/commit/1f1a880c0da4acabb52c877131445b78fc9bed44), besides a (very important) fix for making sure problems, like this one, related to the git repo aren't swallowed silently [ea351c5](https://github.com/Vandebron/mpyl/commit/ea351c5d7715d89bfaafd33a2acf49d6ea28a224)

Thanks, all clear now. I missed that part of the customizable stages PR then

koeves:
I merged your PR and used a new version, but the single commit issue is still not fixed: https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline/job/PR-10859/4
you can see that it still does: 
```
DEBUG    git.cmd  Popen(['git', 'rev-list', 'master..HEAD', '--'], cwd=/home/jenkins/agent/workspace/MPyL_Pipeline_PR-10859, universal_newlines=False, shell=None, istream=None)             
           DEBUG    git.cmd  Popen(['git', 'diff', '--name-only', '6756b00c85abd06ee629b9b30161cd8a5883eed6..6756b00c85abd06ee629b9b30161cd8a5883eed6']
```

i.e. getting the diff between the same revision 6756b00c85abd06ee629b9b30161cd8a5883eed6 (the first commit in my PR: https://github.com/Vandebron/Vandebron/pull/10859)

Also looking at the commit you linked, that seems to solve the empty branch issue but not the single commit one right? Since it checks if the revisions are None, not if it contains only one revision. I'm reopening this PR and I'll try another fix

koeves:
New version works as well: https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline/job/PR-10859/6

src/mpyl/schema/run_properties.schema.yml
```diff
@@ -54,8 +54,8 @@ properties:
         properties:
           deploy_target:
             description: The deploy target.
-            type: ["string", "null"]
-            enum: [ null,  'PullRequest', 'PullRequestBase', 'Production' ]
+            type: [ "string", "null" ]
+            enum: [ null,  'PullRequest', 'PullRequestBase', 'Acceptance', 'Production' ]
```
koeves, line 58: @SamTheisens Acceptance was taken out of the list in https://github.com/Vandebron/mpyl/commit/bf7e9f0535b4111288633daaf819976b3db9f133, I assume by accident?

koeves: I should be more careful with reviewing I guess :D

src/mpyl/schema/run_properties.schema.yml
```diff
@@ -1,29 +1,29 @@
 $schema: http://json-schema.org/draft-07/schema#
 $id: schema/run_properties.schema.yml
 required:
-  - build
+- build
```
Jorg88, line 4: You seem to be doing this change a lot, should it be a formatting or linting rule if possible?

Jorg88 (approved):

src/mpyl/schema/run_properties.schema.yml
koeves: Yeah seems like my yaml formatter is different in pycharm. I'll try to disable it and revert these changes

src/mpyl/steps/models.py
```diff
@@ -125,7 +125,7 @@ def from_configuration(run_properties: Dict, config: Dict):
             details=RunContext.from_configuration(build["run"]),
             target=Target(
                 build["parameters"].get("deploy_target", None)
-                or Target.PULL_REQUEST.value
+                or Target.PULL_REQUEST.value  # pylint: disable=E1101
```
Jorg88, line 128: These warnings are actually blocking all pr's, just ran into them in my pr after rebasing. I'll add a disable I guess, but the warnings actually seem valid when I look at the code? Maybe you can explain it to me when you're back

Jorg88: I fixed these in my cypress pr as well, using the `no-member` as Sam suggested. We do need to look into why these errors don't show up locally..

koeves:
> If there is only one revision, `{revisions[0].hexsha}..{revisions[-1].hexsha}` are identical. So that should work. Did you try it out? Regarding https://github.com/Vandebron/mpyl/pull/155/files#diff-f2e93ed09e3ea127c21228d5487fed442fe1b3364117f6a7ec4705cd660a72a0R452, I think we should avoid using lint error codes because they're not readable. In this case we can replace `E1101` with `no-member` I think.

Exactly what I mean, git is trying to get the difference between the same two revisions (e.g. `git diff first_revision..first_revision`) which is empty! So if there's only one commit, mpyl detects no changes. I have reproduced this, and others have reported this as well.

You wrote above that with this approach we do not need to have master checked out. Is that a realistic use case we should cover? Also when we get the revisions in the same function we do `self._repo.iter_commits(f"{self._config.main_branch}..HEAD")` which already needs to have master.

I think we do need master checked out to correctly get the changes in a branch, otherwise what do you compare to? And also, master _is_ pulled by mpyl if it is not already there anyway.

So from my perspective I'd use the fix in this PR, or go back to the MPL version of doing just `git diff origin/master..`. 

Let me know what you think!

SamTheisens:
Yes, makes sense. Doing `git diff origin/master...` gave problems after rebase IIRC. Don't remember exactly what, unfortunately.

koeves:
> Yes, makes sense. Doing `git diff origin/master...` gave problems after rebase IIRC. Don't remember exactly what, unfortunately.

Thanks, then I'd merge the fix in this PR and see how it holds up over time

PR #162, opened 2023-07-14
Feature/tech 474 extend namespace replace functionality

Jorg88:
----
### 📕 [TECH-474](https://vandebron.atlassian.net/browse/TECH-474) Fix namespace interpolation jorgpost@vandebron.nl 
See [p1687335167697939](https://vandebron.slack.com/archives/CGJDDBETU/p1687335167697939) 

🏗️ Build [3](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-162/3/display/redirect) ✅ Successful, started by _Jorg Post_  
🚀 *[cloudfront-service](https://cloudfront-service-162.test.nl/)*, *job*, *kong-sync*, *[nodeservice](https://nodeservice-162.test.nl/)*, *[sbtservice](https://sbtservice-162.test.nl/)*, *sparkJob*  

[TECH-474]: https://vandebron.atlassian.net/browse/TECH-474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -8,3 +8,4 @@
 /dataSources.local.xml
 /misc.xml
 /libraries/
+watcherTasks.xml
--- a/src/mpyl/steps/deploy/k8s/__init__.py
+++ b/src/mpyl/steps/deploy/k8s/__init__.py
@@ -84,7 +84,7 @@ def substitute_namespaces(
     env = env_vars.copy()
 
     def has_fallback_namespace(env_value: str):
-        regex = re.compile("{namespace=(.*?)}")
+        regex = re.compile("{namespace:(.*?)}")
         return regex.search(env_value)
 
     def get_namespace_for_linked_project(env_value: str, project_name: ProjectName):
@@ -100,7 +100,7 @@ def substitute_namespaces(
         namespace = get_namespace_for_linked_project(env_value, project)
         search_value = project_name + r"\.{namespace}"
         if has_fallback_namespace(env_value):
-            search_value = project_name + r"\.{namespace=.*?}"
+            search_value = project_name + r"\.{namespace:.*?}"
         replace_value = project_name + "." + namespace
         return re.sub(search_value, replace_value, env_value)
 
--- a/src/mpyl/steps/deploy/k8s/__init__.py
+++ b/src/mpyl/steps/deploy/k8s/__init__.py
@@ -83,23 +83,24 @@ def substitute_namespaces(
 ) -> dict[str, str]:
     env = env_vars.copy()
 
-    def is_subproject_within_namespace(env_value: str):
-        regex = re.compile("{namespace/.*?}")
+    def has_fallback_namespace(env_value: str):
+        regex = re.compile("{namespace=(.*?)}")
         return regex.search(env_value)
 
     def get_namespace_for_linked_project(env_value: str, project_name: ProjectName):
         is_part_of_same_deploy_set = project_name in projects_to_deploy
-        if (
-            is_part_of_same_deploy_set or is_subproject_within_namespace(env_value)
-        ) and pr_identifier:
+        if is_part_of_same_deploy_set and pr_identifier:
             return f"pr-{pr_identifier}"
+        fallback_namespace = has_fallback_namespace(env_value)
+        if fallback_namespace:
+            return fallback_namespace.group(1)
         return project_name.namespace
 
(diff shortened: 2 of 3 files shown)


koeves (approved):
Looks good, was just wondering if we should use `:` as the fallback separator? Like we do for env vars

SamTheisens:
@Jorg88 it's not clear to me what purpose of this fallback namespace is. We didn't have this in MPL IIRC? Maybe you could explain the substitution logic in the (pdoc) documentation?

koeves:
> @Jorg88 it's not clear to me what purpose of this fallback namespace is. We didn't have this in MPL IIRC? Maybe you could explain the substitution logic in the (pdoc) documentation?

This was Dan's original message:

```
We recently created a set of services (7 in total) called FlexGW and we use a namespace flex-gateway as recommended in the wiki [here](https://vandebron.atlassian.net/wiki/spaces/DIG/pages/95581326344239/Namespace+reference+usage)
We now need to update flexiot (frontend) to use the FlexGW as a backend. Trouble is, flexiot lives in the batterypack namespace
So for the urls, we define them like this in flexiot project.yml:

- key: CHARGE_POINTS_API_LOCATION
  all: 'http://flexgwchargepoints.flex-gateway.svc.cluster.local:8203/'

Normally for the PR key, we would use http://flexgwchargepoints/.{namespace}.svc.cluster.local:8203
But because flexiot lives in the batterypack namespace, when flexgwchargepoints in unchanged in the PR, it becomes http://flexgwchargepoints.batterypack.svc.cluster.local:8203/, which is wrong.
The relevant code in Mpyl seems to be [here](https://github.com/Vandebron/mpyl/blob/main/src/mpyl/steps/deploy/k8s/__init__.py#L61). As you can see, it assumes that everything is under the same namespace.
Are there any plans to address this?
Thanks!
```

I think it came out because we're starting to adopt the "namespace per service" and not the "namespace per team" approach in k8s. So if we want to interpolate a project that is not part of the same deploy set we would substitute the same namespace as the project yaml of the current project. However if that service lives in a different namespace, this approach would fail. So we can provide a fallback for the interpolation

SamTheisens:
Let's see if we're on the same page. Regardless of the current logic, there are three possibilities for the namespace in which a project will be deployed.

1. the name of the service. This is the default scenario. You get this by leaving the `deployment.namespace` field empty.
2. as specified in `deployment.namespace`. For example, if you have a service and a related database, you can group them together by setting this field.
3. the name of the pullrequest. Any service with changes will get the pr namespaces. Services related to the service (because they have a deployment dependency) with the changes will also deployed in the pr namespace.

Do you agree?

koeves:
> Let's see if we're on the same page. Regardless of the current logic, there are three possibilities for the namespace in which a project will be deployed.
> 
> 1. the name of the service. This is the default scenario. You get this by leaving the `deployment.namespace` field empty.
> 2. as specified in `deployment.namespace`. For example, if you have a service and a related database, you can group them together by setting this field.
> 3. the name of the pullrequest. Any service with changes will get the pr namespaces. Services related to the service (because they have a deployment dependency) with the changes will also deployed in the pr namespace.
> 
> Do you agree?

So if I understand you correctly, your suggestion for Dan's issue is to have a deployment dependency on the FlexGW service in the flexiot frontend's project yaml?

SamTheisens:
Yeah, if I understand correctly flexiot's frontend acts as a gateway to other services. So I'm guessing the usecase is similar to what we do with `nginx-gateway` in front of the `/apps`.
This is a tricky but important bit, so I'd be happy to talk over Whatsapp if you want.

kiequoo:
This is not just about FlexIOT / flexGW. The reason I bring this up in because the platform would like each service to use it's own namespace, rather than just `vdb-services` or `batterypack` or whatever. see the confluence page [here](https://vandebron.atlassian.net/wiki/spaces/DIG/pages/95581326344239/Namespace+reference+usage)
For Sam if you no longer have access, the main point is:
> Use a namespace per service or application (not depending on team structures)

If we really move to this model then adding a deployment dependency to each project would mean a PR deployment of a change in one service could potentially deploy 10x other services for no reason, wasting lots of resources in k8s and being very slow (at least by my understanding)

koeves:
Taken over by #167

PR #163, opened 2023-07-16
Prepare MPyL for use in GHA

SamTheisens:
Improve usability of MPyL for use in Github Actions and/or ArgoCD

 - [x] Dyanmically determine package root. To simplify use of mpyl in scripts.
 - [x] Improve console config: centralized and configurable
 - [x] Export HELM template renderings to file, for debugging or as input to ArgoCD
 - [x] Add (registry) caching support for docker builds
 - [x] Improve documentation around deploy steps

----
 # ⚠️ Could not find ticket corresponding to `feature/TECH-XXX-github-actions-as-runner. Does your branch name follow the correct pattern?


Jorg88 (approved):
There are some nice improvements in this pr that we could use either way, but do we want to merge any GHA specific stuff before we actually decided to use/support GHA?

SamTheisens:
@Jorg AFAIK there is no GHA specific logic in this PR?
Re https://github.com/Vandebron/mpyl/pull/163#discussion_r1271948292, this was a leftover from an earlier attempt. I've replaced it with a warning when no plugins could be found https://github.com/Vandebron/mpyl/pull/163/commits/863ced83cea0b7cb65f8361839d602596a734bc2

src/mpyl/cli/commands/build/mpyl.py
```diff
@@ -126,9 +116,7 @@ def run_mpyl(
                 steps = Steps(
                     logger=logger,
                     properties=mpyl_run_parameters.run_config.run_properties,
-                    steps_collection=StepsCollection(
-                        logger=logger, base_path="src" if params.local else None
-                    ),
+                    steps_collection=StepsCollection(logger=logger),
```
koeves, line 119: We needed this change with Jorg to run mpyl locally. Same idea as the imports in the runner where we check if it's running locally or not. We probably have different local setup then?

src/mpyl/schema/mpyl_config.schema.yml
```diff
@@ -82,6 +82,10 @@ definitions:
     type: object
     additionalProperties: false
     properties:
+      simulate:
+        type: boolean
+        description: 'Simulate kubernetes deployments'
+        default: false
```
koeves, line 88: How is this different from a dry run?

SamTheisens:
@koeves 
 - https://github.com/Vandebron/mpyl/pull/163/files#r1272104330. Are you saying the code in this PR requires changes for local MPyL builds to work on @Jorg88's machine? 
 - https://github.com/Vandebron/mpyl/pull/163#discussion_r1272123548. `dry_run` is a global parameter to the entire run. It can mean different things to different steps. For the build step it means that we won't push the built image to the registry. For HELM deploy it means passing the `--dry-run` to the helm command. That's different from rendering the HELM templates. I've renamed the parameter and added some description to hopefully make the distinction clearer https://github.com/Vandebron/mpyl/pull/163/commits/0dd6dc2e4d50b64e4f54b2f6202637d442f56914. 
   In the longer run, perhaps we should get rid of the global `dry_run` parameter as it's too ambiguous and can't capture all the different behaviors we need per step?

koeves:
> * https://github.com/Vandebron/mpyl/pull/163/files#r1272104330. Are you saying the code in this PR requires changes for local MPyL builds to work on @Jorg88's machine?

Both for him and me, we needed this change to run mpyl locally

SamTheisens:
@koeves it shouldn't be necessary anymore. We just try both locations (with and without src prefix) for finding plugins. Did you try running `163.1927`?

koeves:
> @koeves it shouldn't be necessary anymore. We just try both locations (with and without src prefix) for finding plugins. Did you try running `163.1927`?

Nice, seems to work indeed!

src/mpyl/steps/collection.py
```diff
@@ -11,9 +12,25 @@
...
+        package = "mpyl.steps"
+        local_package = f"src.{package}"
+        default_location, alternative_location = (
+            (local_package, package)
+            if importlib.util.find_spec("src")
+            else (package, local_package)
+        )
+        location = default_location
+
+        self.__load_steps_in_module(".", location)
```
Jorg88, line 26: Isn't the above logic redundant? You can just try the `local_package` first and then `package` on line 28?

src/mpyl/utilities/docker/__init__.py
```diff
@@ -74,7 +92,7 @@ def execute_with_stream(
 ):
     if multiprocess:  # Logger settings need to be re-applied in each process
         logger.setLevel(logging.INFO)
-        logger.addHandler(RichHandler())
+        logger.handlers.clear()
```
Jorg88, line 95: Why this change? Did you check the difference of the logging output of the cypress step (that's the one that uses multiprocessing)?

SamTheisens: Yeah, I checked the output and the problem was that there was a nested logger, which messed up the output. Did you see this too?
Should have taken a screenshot...
I guess you can also just kick off a build and check the logs.

Jorg88: Ah that would explain a lot! Where did you fix the nested logger in this pr? I didn't notice it during my review

src/mpyl/steps/collection.py
SamTheisens: The other location is tried if the default location has failed (no plugins could be loaded). 
But I guess I could improve the naming to make that a bit clearer?

src/mpyl/utilities/docker/__init__.py
SamTheisens: This _is_ the fix for the nesting problem. I think `addHandler` was the culprit.

src/mpyl/steps/collection.py
Jorg88: I'm saying that unless `if importlib.util.find_spec("src")` is important, you don't need `default_location`& `alternative_location` and you can just try to load `local_package` and fallback to `package` otherwise.

src/mpyl/utilities/docker/__init__.py
Jorg88: Oh... it's been a while, but I swear this was needed to get any sort of formatted logging from the cypress step. Anyway, I'm happy if it works now

src/mpyl/steps/collection.py
SamTheisens: Ah! I see what you mean. Yeah, this does seem double. But the problem I ran into was that `importlib.util.find_spec("src")` can return something if you use the mpyl _library_ in a local script in the mpyl repo. Or any other repo that has a `src` folder for that matter.
So it's a bit of a corner case. Ultimately the proof of the pudding is in what results in plugins being loaded. The check `find_spec` helps us look in the most likely location first.

Jorg88 (approved):

koeves (approved):
Do you think it's possible to enable registry caching for the sbt build step as well?

SamTheisens:
> Do you think it's possible to enable registry caching for the sbt build step as well?

Unfortunately not I'm afraid. The compiled scala code is kept on the filesystem scattered over `/target/` folders, not in docker image layers. Only at the very end a docker image is wrapped around the fat jar. 
But then again, there aren't any stages that would be worth caching either, like the `yarn install` that you get with frontend builds. 
We have the MPyL caching of completed artifacts anyway. This gets lost accross build executors, but that's a more general problem for which we could find a general solution. Is there a ticket for that already?

koeves:
> > Do you think it's possible to enable registry caching for the sbt build step as well?
> 
> Unfortunately not I'm afraid. The compiled scala code is kept on the filesystem scattered over `/target/` folders, not in docker image layers. Only at the very end a docker image is wrapped around the fat jar. But then again, there aren't any stages that would be worth caching either, like the `yarn install` that you get with frontend builds. We have the MPyL caching of completed artifacts anyway. This gets lost accross build executors, but that's a more general problem for which we could find a general solution. Is there a ticket for that already?

Made one this morning! https://vandebron.atlassian.net/browse/TECH-521
We were talking about this as well, for tag deployments we could look into the registry to see if an image with that certain tag exists and then we can skip the build steps? Not sure if we can figure something out for the PR images

SamTheisens:
> > > Do you think it's possible to enable registry caching for the sbt build step as well?
> > 
> > 
> > Unfortunately not I'm afraid. The compiled scala code is kept on the filesystem scattered over `/target/` folders, not in docker image layers. Only at the very end a docker image is wrapped around the fat jar. But then again, there aren't any stages that would be worth caching either, like the `yarn install` that you get with frontend builds. We have the MPyL caching of completed artifacts anyway. This gets lost accross build executors, but that's a more general problem for which we could find a general solution. Is there a ticket for that already?
> 
> Made one this morning! https://vandebron.atlassian.net/browse/TECH-521 We were talking about this as well, for tag deployments we could look into the registry to see if an image with that certain tag exists and then we can skip the build steps? Not sure if we can figure something out for the PR images

👍

If we commit the `.mpl` folders to a repo at the end of a build, and pull them at the start, we effectively store the entire build state. It will allow us to transfer the state to a new executor and will automatically skip build steps. Without any "fiddling" in the docker registry.
Fits with gitops principles and naturally paves the way for ArgoCD.
WDYT?

Jorg88:
> > > > Do you think it's possible to enable registry caching for the sbt build step as well?
> > > 
> > > 
> > > Unfortunately not I'm afraid. The compiled scala code is kept on the filesystem scattered over `/target/` folders, not in docker image layers. Only at the very end a docker image is wrapped around the fat jar. But then again, there aren't any stages that would be worth caching either, like the `yarn install` that you get with frontend builds. We have the MPyL caching of completed artifacts anyway. This gets lost accross build executors, but that's a more general problem for which we could find a general solution. Is there a ticket for that already?
> > 
> > 
> > Made one this morning! https://vandebron.atlassian.net/browse/TECH-521 We were talking about this as well, for tag deployments we could look into the registry to see if an image with that certain tag exists and then we can skip the build steps? Not sure if we can figure something out for the PR images
> 
> 👍
> 
> If we commit the `.mpl` folders to a repo at the end of a build, and pull them at the start, we effectively store the entire build state. It will allow us to transfer the state to a new executor and will automatically skip build steps. Without any "fiddling" in the docker registry. Fits with gitops principles and naturally paves the way for ArgoCD. WDYT?

We had the same idea yesterday, but why a git repo, isn't something like a s3 bucket more suited for a task like this?

SamTheisens:
> > > > > Do you think it's possible to enable registry caching for the sbt build step as well?
> > > > 
> > > > 
> > > > Unfortunately not I'm afraid. The compiled scala code is kept on the filesystem scattered over `/target/` folders, not in docker image layers. Only at the very end a docker image is wrapped around the fat jar. But then again, there aren't any stages that would be worth caching either, like the `yarn install` that you get with frontend builds. We have the MPyL caching of completed artifacts anyway. This gets lost accross build executors, but that's a more general problem for which we could find a general solution. Is there a ticket for that already?
> > > 
> > > 
> > > Made one this morning! https://vandebron.atlassian.net/browse/TECH-521 We were talking about this as well, for tag deployments we could look into the registry to see if an image with that certain tag exists and then we can skip the build steps? Not sure if we can figure something out for the PR images
> > 
> > 
> > 👍
> > If we commit the `.mpl` folders to a repo at the end of a build, and pull them at the start, we effectively store the entire build state. It will allow us to transfer the state to a new executor and will automatically skip build steps. Without any "fiddling" in the docker registry. Fits with gitops principles and naturally paves the way for ArgoCD. WDYT?
> 
> We had the same idea yesterday, but why a git repo, isn't something like a s3 bucket more suited for a task like this?

For ArgoCD you need your deployment metadata to be in a repo anyway.
And you get traceability, immutability, blockchain ;-), etc. for free. We also already have all the machinery in place to talk to git repos.

PR #165, opened 2023-07-19
Feature/tech 484 add support for normal secrets

Jorg88:
----
### 📕 [TECH-484](https://vandebron.atlassian.net/browse/TECH-484) Add support for normal secrets danielkoves@vandebron.nl 

🏗️ Build [8](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-165/8/display/redirect) ✅ Successful, started by _Daniel Koves_  
🚀 *[cloudfront-service](https://cloudfront-service-165.test.nl/)*, *job*, *kong-sync*, *[nodeservice](https://nodeservice-165.test.nl/)*, *[sbtservice](https://sbtservice-165.test.nl/)*, *sparkJob*  

[TECH-484]: https://vandebron.atlassian.net/browse/TECH-484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/mpyl/steps/deploy/k8s/chart.py
```diff
@@ -481,6 +485,23 @@ def _get_job(self) -> Job:
...
+                    name=e.key,
+                    value_from=V1EnvVarSource(
+                        secret_key_ref=V1SecretKeySelector(
+                            key=e.key, name=self.release_name, optional=False
+                        )
+                    ),
+                ),
+                secret_list,
+            )
+        )
```
koeves, line 503: I like this style better than`list(map(...))` but leave it up to you :)
```suggestion
        return [
            V1EnvVar(
                name=e.key,
                value_from=V1EnvVarSource(
                    secret_key_ref=V1SecretKeySelector(
                        key=e.key, name=self.release_name, optional=False
                    )
                ),
            )
            for e in secret_list
        ]
```

koeves (approved):

koeves:
Not sure if the current approach works..

I tried it with my project (https://github.com/Vandebron/Vandebron/blob/c8908294830ae0eaac13b0047b5e63a9c5339c41/energy-trading/cbp/cbp-bids-manager/deployment/project.yml#L67-L72) and it's outputting 

```yaml
- name: POSTGRES_PASSWORD
              valueFrom:
                secretKeyRef:
                  key: POSTGRES_PASSWORD
                  name: cbpbidsmanager
                  optional: false
```

instead of

```yaml
- name: POSTGRES_PASSWORD
              valueFrom:
                secretKeyRef:
                  key: password
                  name: trading-cbp-manager-postgres-vdb-user
                  optional: false
```

It's because we use `self.release_name` in `src/mpyl/steps/deploy/k8s/chart.py:496`

koeves (changes requested):
We need to parse correctly the kubernetes part of the yaml, based on the schema:

```yaml
          kubernetes:
            type: array
            items:
              type: object
              additionalProperties: false
              required:
                - key
                - valueFrom
              properties:
                key:
                  type: string
                valueFrom:
                  oneOf:
                    - $ref: '#/definitions/fieldRef'
                    - $ref: '#/definitions/secretKeyRef'
                    - $ref: '#/definitions/resourceFieldRef'               
```

Jorg88:
> Not sure if the current approach works..
> 
> I tried it with my project (https://github.com/Vandebron/Vandebron/blob/c8908294830ae0eaac13b0047b5e63a9c5339c41/energy-trading/cbp/cbp-bids-manager/deployment/project.yml#L67-L72) and it's outputting
> 
> ```yaml
> - name: POSTGRES_PASSWORD
>               valueFrom:
>                 secretKeyRef:
>                   key: POSTGRES_PASSWORD
>                   name: cbpbidsmanager
>                   optional: false
> ```
> 
> instead of
> 
> ```yaml
> - name: POSTGRES_PASSWORD
>               valueFrom:
>                 secretKeyRef:
>                   key: password
>                   name: trading-cbp-manager-postgres-vdb-user
>                   optional: false
> ```
> 
> It's because we use `self.release_name` in `src/mpyl/steps/deploy/k8s/chart.py:496`
> 
> We should probably use something other than `KeyValueProperty` for this

This code isn't new though, it's already being used for sealed secrets as well. In the end, these things give the same result in the container, which is a normal secret. Are you saying that logic is also wrong? If so - I can change it, but it would be weird we didn't find issues yet with running containers. If not - I don't understand what is happening 😬

Jorg88:
> We need to parse correctly the kubernetes part of the yaml, based on the schema:
> 
> ```yaml
>           kubernetes:
>             type: array
>             items:
>               type: object
>               additionalProperties: false
>               required:
>                 - key
>                 - valueFrom
>               properties:
>                 key:
>                   type: string
>                 valueFrom:
>                   oneOf:
>                     - $ref: '#/definitions/fieldRef'
>                     - $ref: '#/definitions/secretKeyRef'
>                     - $ref: '#/definitions/resourceFieldRef'               
> ```

That's the input right, we already are doing so?

koeves:
Sealed secrets have the same structure as the environment variables in the project yaml:
```yaml
      - key: SOME_ENV
        pr: "PullRequest"
        test: "Test"
        acceptance: "Acceptance"
        production: "Production"
```

this is what the `KeyValuePropertiy` parses

whereas these secrets have this form:

```yaml
      - key: SOME_SECRET_ENV
        valueFrom:
          secretKeyRef:
            name: SOME_SECRET_ENV
            key: sbt-job
            optional: false
```

so we need to parse a different data type, starting with `valueFrom` then `secretKeyRef` etc.

I'll push what I'm thinking of

src/mpyl/project.py
```diff
@@ -111,6 +111,60 @@ def from_config(values: dict):
...
+    key: str
+    optional: bool
+    key_name: str = "secretKeyRef"
+
+    @staticmethod
+    def from_config(values: dict):
+        return SecretKeyRef(
+            name=values.get("name"),
+            key=values.get("key"),
+            optional=values.get("optional", False),
```
Jorg88, line 126: is `optional` optional? 😅

src/mpyl/project.py
```diff
@@ -111,6 +111,60 @@ def from_config(values: dict):
...
+
+    @staticmethod
+    def from_config(values: dict):
+        key = values.get("key")
+        value_from = values.get("valueFrom")
+
+        if not key or not value_from:
+            raise KeyError("KeyValueRef must have a key or valueFrom set")
+
+        key_name = list(value_from.keys())[0]
```
Jorg88, line 153: are we actually sure there is always only 1 key?

src/mpyl/steps/deploy/k8s/chart.py
```diff
@@ -481,6 +486,34 @@ def _get_job(self) -> Job:
...
+                value_from=V1EnvVarSource(
+                    secret_key_ref=V1SecretKeySelector(
+                        key=e.key, name=self.release_name, optional=False
+                    )
+                ),
+            )
+            for e in secret_list
+        ]
+
+    def _create_secret_env_vars(self, secret_list: list[KeyValueRef]) -> list[V1EnvVar]:
```
Jorg88, line 504: there must be a way these functions can be combined

src/mpyl/project.py
koeves: yes, according to the schema:
```yaml
        - name: SOME_SECRET_ENV
          valueFrom:
            secretKeyRef:
              key: password
              name: some-secret-name
              optional: false
        - name: KUBERNETES_NAMESPACE
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
```
so the list of keys are `list(secretKeyRef)` in the first case and `list(fieldRef)` in the second case

src/mpyl/project.py
koeves: No it's actually not :D

src/mpyl/steps/deploy/k8s/chart.py
koeves: I can make an interface to map a KeyValueRef and KeyvValueProperty to V1EnvVar but then this implementation moves to `project.py` which is weird I think

src/mpyl/project.py
Jorg88: which checks?

koeves: if not key or not name or not optional triggers if optional is False even though from a typing perspective it is a Some(False) and not a None

koeves:
Also tetsted it with this PR 
https://github.com/Vandebron/Vandebron/pull/10997
https://jenkins.k8s-serv-backend.vdbinfra.nl/view/MPyL/job/MPyL%20Pipeline/view/change-requests/job/PR-10997/
and we have:

```yaml
- name: POSTGRES_PASSWORD
              valueFrom:
                secretKeyRef:
                  key: password
                  name: trading-cbp-manager-postgres-vdb-user
                  optional: false
```

in the deployment yaml

Jorg88:
Maybe we want to put this ticket on hold until we've discussed https://vandebron.atlassian.net/browse/TECH-486 ?

koeves:
> Maybe we want to put this ticket on hold until we've discussed https://vandebron.atlassian.net/browse/TECH-486 ?

I'd merge this, that ticket is just a usability improvement on this one. If we do find that platform should go with sealed secrets, it will still take some time until that is realised. Until that point, we cannot use mpyl for all projects that mount normal secrets. And even so, it's a feature of mpl/mpyl that I think is useful (mounting secrets), even if it's misused here

koeves (approved):

koeves (approved):

PR #166, opened 2023-07-19
[TECH-440] Implement a new way to get changes in branch that ignores …

Jorg88:
…merge commits

branch: feature/TECH-440-ignore-merge-commits

----
### 📕 [TECH-440](https://vandebron.atlassian.net/browse/TECH-440) Ignore merge commits in build plan jorgpost@vandebron.nl 
Currently your build plan gets messed up if you’ve done a merge instead of a rebase, it will recognize all changes of the merge. It would be nice if we could ignore merge commits somehow while determining the build plan.

🏗️ Build [6](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-166/6/display/redirect) ✅ Successful, started by _Jorg Post_  
🚀 *[cloudfront-service](https://cloudfront-service-166.test.nl/)*, *job*, *kong-sync*, *[nodeservice](https://nodeservice-166.test.nl/)*, *[sbtservice](https://sbtservice-166.test.nl/)*, *sparkJob*  

[TECH-440]: https://vandebron.atlassian.net/browse/TECH-440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/mpyl/utilities/repo/__init__.py
+++ b/src/mpyl/utilities/repo/__init__.py
@@ -135,11 +135,12 @@ class Repository:
         return Revision(count, str(revision), intersection)
 
     def changes_in_branch(self) -> list[Revision]:
+        self.pull_main_branch()
         revisions = list(
             reversed(
                 list(
                     self._repo.iter_commits(
-                        f"origin/{self._config.main_branch}..HEAD", no_merges=True
+                        f"{self._config.main_branch}..HEAD", no_merges=True
                     )
                 )
             )
--- a/src/mpyl/utilities/repo/__init__.py
+++ b/src/mpyl/utilities/repo/__init__.py
@@ -136,7 +136,13 @@ class Repository:
 
     def changes_in_branch(self) -> list[Revision]:
         revisions = list(
-            reversed(list(self._repo.iter_commits(f"{self._config.main_branch}..HEAD")))
+            reversed(
+                list(
+                    self._repo.iter_commits(
+                        f"origin/{self._config.main_branch}..HEAD", no_merges=True
+                    )
+                )
+            )
         )
 
         logging.debug(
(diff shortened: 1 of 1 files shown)


src/mpyl/utilities/repo/__init__.py
```diff
@@ -146,17 +152,16 @@ def changes_in_branch(self) -> list[Revision]:
         if not revisions:
             return []
 
-        first_revision = (
-            revisions[0].hexsha if len(revisions) != 1 else self._config.main_branch
-        )
```
koeves, line 151: We still need this part I think

Jorg88: Yeah I wasn't sure, but in which way though?

koeves: The same I think.. if there's only one commit, we should diff with main_branch and that one revision

Jorg88: It works fine with only one commit, I tested it with a second branch

koeves (approved):

SamTheisens:
Although I'm not a fan of using merge instead of doing a rebase / create a pristine branch, I guess in most cases you don't want to (re)build whatever is already in master.
But what about merges from other branches than master, I guess you's want to build the changes in that?

Two remaining questions:
 - Why is `no_merges=True` not sufficient and does `origin/` need to be added?
 - This https://github.com/Vandebron/mpyl/pull/166/files#diff-6040da3442be1c39607355675768292629b1dbebadd4fd299aca63e131ccab5eR156-R161 can be written more cleanly as a for comprehension I think.

Jorg88:
> Although I'm not a fan of using merge instead of doing a rebase / create a pristine branch, I guess in most cases you don't want to (re)build whatever is already in master. But what about merges from other branches than master, I guess you's want to build the changes in that?
> 
> Two remaining questions:
> 
>     * Why is `no_merges=True` not sufficient and does `origin/` need to be added?
> 
>     * This https://github.com/Vandebron/mpyl/pull/166/files#diff-6040da3442be1c39607355675768292629b1dbebadd4fd299aca63e131ccab5eR156-R161 can be written more cleanly as a for comprehension I think.

Yeah we can still promote the usage of rebasing instead of merging in the company, but I don't think a branch should be 'ruined' (in terms of mpyl) when you did do a merge. 

- I didn't think about the merges from other branches than master, maybe we can discuss that tomorrow during our call?
- adding `origin/` is not related the no merges tag, I added it since there are ways to reach this code in which the latest main/master is not pulled yet. Is there a downside to using it? This way we're sure you're always comparing to the latest version.
- True for the `for` comprehension! I'm clearly still a js/ts dev, will update

SamTheisens:
- I agree that merges with main should not trigger builds for all files touched in them. That should at least not be the default behavior.
 - I wonder what the scenarios are in which master hasn't been pulled. Do you have any idea? Maybe it's better to solve that problem at the source.
 - Yeah, took me a while to get used to Python's for comprehension syntax. But now I think they're quite elegant and avoid having to introduce mutable collections in your code. 
Would be happy to discuss further tomorrow!

SamTheisens (approved):
LGTM 👍

PR #169, opened 2023-07-26
fix/namespace interpolation for tag deployments

koeves:
- Fix PR interpolation when tag is also set
- Make getting change ID more robust by falling back to `gh`

----
 # ⚠️ Could not find ticket corresponding to `fix/namespace_interpolation_for_tag_deployments. Does your branch name follow the correct pattern?

### DIFF
--- a/src/mpyl/steps/models.py
+++ b/src/mpyl/steps/models.py
@@ -111,8 +111,10 @@ class RunProperties:
         build = run_properties["build"]
         versioning_config = build["versioning"]
 
-        pr_num: Optional[int] = get_pr_number(versioning_config.get("pr_number"))
         tag: Optional[str] = versioning_config.get("tag")
+        pr_num: Optional[int] = (
+            None if tag else get_pr_number(versioning_config.get("pr_number"))
+        )
 
         versioning = VersioningProperties(
             revision=versioning_config["revision"],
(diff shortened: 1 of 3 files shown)


src/mpyl/utilities/github/__init__.py
```diff
@@ -58,3 +61,23 @@ def get_pr_for_branch(repo: Repository, branch: str) -> PullRequest:
         )
 
     return pulls.pop()
+
+
+def get_pr_number(pr_number_from_config: Optional[str] = None) -> Optional[int]:
+    try:
+        if pr_number_from_config:
+            return int(pr_number_from_config)
```
koeves, line 69: Move this to models

src/mpyl/utilities/github/__init__.py
```diff
@@ -58,3 +61,23 @@ def get_pr_for_branch(repo: Repository, branch: str) -> PullRequest:
         )
 
     return pulls.pop()
+
+
+def get_pr_number(pr_number_from_config: Optional[str] = None) -> Optional[int]:
```
koeves, line 66:
```suggestion
def get_pr_number_from_gh(pr_number_from_config: Optional[str] = None) -> Optional[int]:
```

src/mpyl/steps/models.py
```diff
@@ -110,13 +111,15 @@ def from_configuration(run_properties: Dict, config: Dict):
...
+        tag: Optional[str] = versioning_config.get("tag")
+        pr_num: Optional[int] = (
+            None if tag else get_pr_number(versioning_config.get("pr_number"))
+        )
 
         versioning = VersioningProperties(
             revision=versioning_config["revision"],
             branch=versioning_config["branch"],
-            pr_number=int(pr_num) if pr_num else None,
+            pr_number=pr_num if pr_num else None,
```
koeves, line 122: Not needed

PR #175, opened 2023-08-02
[TECH-490] Upgrade python to v3.11

Jorg88:
branch: feature/TECH-490-migrate-mpyl-to-python-3.11

I'm actually not sure if anything else needs to be done for this @koeves ?


Pipfile
```diff
@@ -34,6 +34,7 @@ twine = "==4.0.1"
 types-six = "==1.16.21.4"
 dagit = "==1.2.6"
 dagster = "==1.2.6"
+dill = "==0.3.7"
```
SamTheisens, line 37: why's this?

Pipfile.lock
```diff
@@ -1654,11 +1651,11 @@
...
             "hashes": [
-                "sha256:a07ffd2351b8c678dfc4a856a3005f8067aea51d6ba6c700796a4d9e280f39f0",
-                "sha256:e5db55f3687856d8fbdab002ed78544e1c4559a130302693d839dfe8f93f2373"
+                "sha256:76b122c08ef4ce2eedcd4d1abd8e641114bfc6c2867f49f3c41facf65bf19f5e",
+                "sha256:cc1c8b182eb3013e24bd475ff2e9295af86c1a38eb1aff128dac8962a9ce3c03"
             ],
-            "markers": "python_version < '3.11'",
-            "version": "==0.3.6"
+            "index": "pypi",
+            "version": "==0.3.7"
```
SamTheisens, line 1658: I get:
```
            "markers": "python_version >= '3.11'",
            "version": "==0.3.7"
```
if I run `pipenv install -d` on python 3.11.4

SamTheisens (approved):
Looks good to me. Should we add/update info about the python version to use in the dev section of the documentation?

Pipfile.lock
Jorg88: I'm on the same version and I ran the same cmd, I'll never fully understand python 😅

Pipfile
Jorg88: Because pylint doesn't work without it, if you know a better solution please let me know. The only other thing I could do was changing the lock file manually, but that didn't seem like the right approach.

.github/workflows/build-package.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/setup-python@v4
         id: cache-dependencies
         with:
-          python-version: '3.9'
+          python-version: '3.11'
```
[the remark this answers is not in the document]
Jorg88: I guess we could, but why? 😬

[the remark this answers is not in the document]
SamTheisens: Because it's cleaner?

Pipfile
SamTheisens: 🤔 curious. The dill dependency was already present in the `pipfile.lock`. Pylint worked with 3.11 on my machine.

Pipfile.lock
SamTheisens: Did you create a clean pipenv? That's what I did.

Pipfile
Jorg88: It also worked on my machine, but not in the github action

SamTheisens: Interesting. Maybe you can explain that in a comment with the depedency

SamTheisens:
I tested the release as well. Seems to work with either python version.
This reminds me of something. I think it would be good to bundle typing info with the egg. Maybe you could create a ticket for that?

.github/workflows/build-package.yml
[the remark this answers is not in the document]
Jorg88: Ah that's what you meant! I thought you wanted to make it dynamic at first

Pipfile.lock
Jorg88: Okay I assumed the python version in `requires` worked like in node where you can just specify the typescript version and it will handle everything. Seems I was wrong, remade a new venv with 3.11, deleted the lock file and ran `pipenv update` again, now everything should be the same as yours!

Pipfile
Jorg88: This is also fixed with the change below

SamTheisens:
@Jorg88 is there a blocking issue that prevents us from merging this?

koeves:
> @Jorg88 is there a blocking issue that prevents us from merging this?

We needed to find a way to properly update and test the jenkins-test-dind image. Because we need to update it otherwise python 3.11 is not available. We synced with platform, agreed that we should create a new image, test it in the test jenkins deployment, and then we can merge it. I think we just didn't get around to doing that

PR #196, opened 2023-08-28
[TECH-477] Updating AWS credential requirements

benjaminlynch100:
Making the access and secret keys optional in the config as it's very common to use AWS credential or config files for authentication. Boto3 will automatically look for these files if nothing is passed to the client.

branch:feature/TECH-477-aws-credentials

local build reading from credentials file:
Screenshot 2023-08-28 at 11 39 34

----
### 📕 [TECH-477](https://vandebron.atlassian.net/browse/TECH-477) Read AWS credentials from credentials file instead of specifying env vars benjaminlynch@vandebron.nl 
Use `boto3}} or read manually from {{~/.aws/credentials` if there are no AWS environment variables

github token is handled at: `mpyl.cli.commands.build.jenkins.get_token` as an example

🏗️ Build [19](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-196/19/display/redirect) ✅ Successful, started by _Ben Lynch_  
🚀 *[cloudfront-service](https://cloudfront-service-196.test.nl/)*, *job*, *kong-sync*, *[nodeservice](https://nodeservice-196.test.nl/)*, *[sbtservice](https://sbtservice-196.test.nl/)*, *sparkJob*  

[TECH-477]: https://vandebron.atlassian.net/browse/TECH-477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/mpyl/project.py
+++ b/src/mpyl/project.py
@@ -347,10 +347,14 @@ class Traefik:
 @dataclass(frozen=True)
 class S3Bucket:
     bucket: TargetProperty[str]
+    region: str
 
     @staticmethod
     def from_config(values: dict):
-        return S3Bucket(bucket=TargetProperty.from_config(values.get("bucket", {})))
+        return S3Bucket(
+            bucket=TargetProperty.from_config(values.get("bucket", {})),
+            region=values.get("region", {}),
+        )
 
 
 @dataclass(frozen=True)
--- a/src/mpyl/schema/mpyl_config.schema.yml
+++ b/src/mpyl/schema/mpyl_config.schema.yml
@@ -363,6 +363,3 @@ definitions:
         type: string
       secretAccessKey:
         type: string
-    required:
-      - accessKeyId
-      - secretAccessKey
--- a/src/mpyl/schema/mpyl_config.schema.yml
+++ b/src/mpyl/schema/mpyl_config.schema.yml
@@ -360,6 +360,9 @@ definitions:
     type: object
     properties:
       accessKeyId:
-        type: string
+        type: [ string, null ]
       secretAccessKey:
-        type: string
+        type: [ string, null ]
+    dependencies:
+      accessKeyId: [ secretAccessKey ]
+      secretAccessKey: [ accessKeyId ]
--- a/src/mpyl/utilities/s3/__init__.py
+++ b/src/mpyl/utilities/s3/__init__.py
@@ -15,17 +15,18 @@ from ...steps.models import RunProperties
 
 @dataclass
 class S3ClientConfig:
-    service_name = "s3"
-    region_name = "eu-central-1"
-    access_key_id: Optional[str]
-    secret_access_key: Optional[str]
     bucket_name: str
     bucket_root_path: str
+    service_name = "s3"
+    region_name = "eu-central-1"
+    access_key_id: Optional[str] = None
+    secret_access_key: Optional[str] = None
 
     def __init__(self, run_properties: RunProperties, bucket_name: str):
-        s3_connection_info = run_properties.config.get("s3", None)
-        self.access_key_id = s3_connection_info.get("accessKeyId")
-        self.secret_access_key = s3_connection_info.get("secretAccessKey")
+        s3_connection_info = run_properties.config.get("s3")
+        if s3_connection_info is not None:
+            self.access_key_id = s3_connection_info.get("accessKeyId")
+            self.secret_access_key = s3_connection_info.get("secretAccessKey")
         self.bucket_name = bucket_name
         self.bucket_root_path = run_properties.versioning.identifier
 
@@ -34,6 +35,10 @@ class S3Client:
     def __init__(self, logger: Logger, config: S3ClientConfig):
         """
         Creates a client that provides a wrapper for uploading a directory to an S3 bucket
+
+        Credentials can be provided via the s3.accessKeyId and s3.secretAccessKey properties in the mpyl_config or
+        via the numerous ways supported by boto3 (see
+        https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html)
 
         :param config: the S3 client config containing the necessary credentials and bucket information
         """
--- a/src/mpyl/utilities/s3/__init__.py
+++ b/src/mpyl/utilities/s3/__init__.py
@@ -22,12 +22,15 @@ class S3ClientConfig:
     access_key_id: Optional[str] = None
     secret_access_key: Optional[str] = None
 
-    def __init__(self, run_properties: RunProperties, bucket_name: str):
+    def __init__(
+        self, run_properties: RunProperties, bucket_name: str, bucket_region: str
+    ):
         s3_connection_info = run_properties.config.get("s3")
         if s3_connection_info is not None:
             self.access_key_id = s3_connection_info.get("accessKeyId")
             self.secret_access_key = s3_connection_info.get("secretAccessKey")
         self.bucket_name = bucket_name
+        self.region_name = bucket_region
         self.bucket_root_path = run_properties.versioning.identifier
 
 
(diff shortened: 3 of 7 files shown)


src/mpyl/utilities/s3/__init__.py
```diff
@@ -15,17 +15,18 @@
 
 @dataclass
 class S3ClientConfig:
-    service_name = "s3"
-    region_name = "eu-central-1"
-    access_key_id: Optional[str]
-    secret_access_key: Optional[str]
     bucket_name: str
     bucket_root_path: str
+    service_name: str = field(default="s3")
```
Jorg88, line 20: why do we need `field` compared to just `service_name: str = "s3"` ?

src/mpyl/utilities/s3/__init__.py
```diff
@@ -15,17 +15,18 @@
...
+    region_name: str = field(default="eu-central-1")
+    access_key_id: Optional[str] = field(default=None)
+    secret_access_key: Optional[str] = field(default=None)
```
Jorg88, line 23: `default=None` seems weird, isn't `None` already the default by default 😁

src/mpyl/utilities/s3/__init__.py
```diff
@@ -15,17 +15,18 @@
...
 
     def __init__(self, run_properties: RunProperties, bucket_name: str):
         s3_connection_info = run_properties.config.get("s3", None)
-        self.access_key_id = s3_connection_info.get("accessKeyId")
-        self.secret_access_key = s3_connection_info.get("secretAccessKey")
+        if s3_connection_info is not None:
+            self.access_key_id = s3_connection_info.get("accessKeyId", None)
```
Jorg88, line 28: since you removed the 'required' check in the schema, shouldn't we now do a check here that they are filled?

src/mpyl/utilities/s3/__init__.py
benjaminlynch100: we don't, I had just been looking at the dataclass docs and saw it, but it's not necessary - I removed it

src/mpyl/utilities/s3/__init__.py
benjaminlynch100: my idea is that there is no required check anymore. You can define them in the config or you can have any of the local setups for AWS credentials. If you don't define them and don't have the local auth setup then the client creation will throw and exception

src/mpyl/utilities/s3/__init__.py
benjaminlynch100: updated

benjaminlynch100: oh, they actually don't get initialised to None by default, so it has to be specified. I just tested it again locally after removing the default value

src/mpyl/utilities/s3/__init__.py
Jorg88: What do you mean with "you can have any of the local setups for AWS credentials"? I'm just thinking that if the client below is used without these credentials it's probably better to throw our own error beforehand already instead of depending on whatever unauthenticated error we get back

benjaminlynch100: I think it makes sense to rely on Boto3 since it automatically searches for credentials in a very [flexible way](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html).  For us to check in advance we'd have to do the same checks Boto3 already does which seems unnecessary if we can just let it do the checks.

Jorg88 (approved):

src/mpyl/utilities/s3/__init__.py
```diff
@@ -15,17 +15,18 @@
...
 @dataclass
 class S3ClientConfig:
-    service_name = "s3"
-    region_name = "eu-central-1"
-    access_key_id: Optional[str]
-    secret_access_key: Optional[str]
     bucket_name: str
     bucket_root_path: str
+    service_name = "s3"
+    region_name = "eu-central-1"
```
koeves, line 21: Can we move these to the mpyl_config as well?

benjaminlynch100: we can! see the latest commit https://github.com/Vandebron/mpyl/pull/196/commits/3fa126548a20f4ae853fade46e2ff7897e969b46. 

- I left the service name hardcoded as this client is only for s3 operations
- I took the liberty of renaming the cloudfront related properties to be more meaningful
- I also updated how the object path is constructed as it wasn't working to well locally

src/mpyl/utilities/s3/__init__.py
```diff
@@ -114,8 +134,8 @@ def upload_file(self, src_path: str, dst_path: str):
         :param src_path: the local relative path of the file
         :param dst_path: the destination path within the bucket
         """
-        self._logger.debug(f"Uploading to bucket: {self._buck_name}")
-        self._logger.debug(f"Uploading file at '{src_path}' to '{dst_path}'")
+        print(f"Uploading to bucket: {self._buck_name}")
+        print(f"Uploading file at '{src_path}' to '{dst_path}'")
```
Jorg88, line 138: I think you forgot to revert these changes

src/mpyl/project.py
```diff
@@ -333,12 +333,16 @@ def from_config(values: dict):
...
+    name: TargetProperty[str]
+    region: str
 
     @staticmethod
     def from_config(values: dict):
-        return S3Bucket(bucket=TargetProperty.from_config(values.get("bucket", {})))
+        return CloudFrontS3Bucket(
+            name=TargetProperty.from_config(values.get("name", {})),
+            region=values.get("region", {}),
+        )
```
SamTheisens, line 345: To what extent is this bucket with specified region specific to Cloudfront?

In general I'm in favor of specific naming, but in this case I would tend toward referring to `s3` here, because 1. I think it's quite likely that this config will be used in other deployment scenarios involving s3 buckets, 2. the cloudfront deploy step already specifies the cloudfront specific use of s3 buckets, 3. MPyL is open source, so a (slightly) higher focus on reusability than with proprietary code would be justified.

src/mpyl/schema/mpyl_config.schema.yml
```diff
@@ -360,9 +360,6 @@ definitions:
...
     properties:
       accessKeyId:
-        type: string
+        type: [ string, null ]
       secretAccessKey:
-        type: string
-    required:
-      - accessKeyId
-      - secretAccessKey
+        type: [ string, null ]
```
SamTheisens, line 365: Are these properties independently optional? If they are dependent, you could specify that in the schema like:
```yaml
    dependencies:
      "accessKeyId": [ secretAccessKey ]
      "secretAccessKey": [ accessKeyId ]
```

SamTheisens (changes requested):
Looks good! I'm surprised this issue wasn't reported earlier, because I presume cloudfront deployment didn't work at all?

src/mpyl/project.py
benjaminlynch100: It's not specific to cloudfront but I was looking at the usages of the s3 config property and it's only being used for cloudfront so I thought it might be better to make that clear in the config.
I can revert the name back, I was just thinking it might be better to keep things specific until we have other known use cases

SamTheisens: If this were purely internal code I would agree, but in this case I would be in favor of reverting the name back to what it was. I don't think it's likely to lead to confusion and it saves us another schema change by the time we create other s3 related steps.

src/mpyl/schema/mpyl_config.schema.yml
benjaminlynch100: yep, they are dependent - I've updated it

src/mpyl/utilities/s3/__init__.py
benjaminlynch100: good catch! fixed

src/mpyl/project.py
benjaminlynch100: I've reverted the naming back to s3

benjaminlynch100:
> Looks good! I'm surprised this issue wasn't reported earlier, because I presume cloudfront deployment didn't work at all?

The Clodufront deploy has been working, it's just that the temporary folder seems to have a different structure when MPyL is run on Jenkins vs locally. When I originally developed it, I tested the s3 upload via Jenkins as it set the credentials and only tested dry runs locally, so I incorrectly assumed a fixed file structure when using the tmp folders

SamTheisens (approved):

ewaldschrader (approved):

PR #197, opened 2023-08-28
Correctly parse resource requests and replica count

koeves:
- Correctly parse instance (replica) numbers from project yaml (it was not under `deployment.kubernetes.resources.limit` but under `deployment.kubernetes.resources`)
- Take into account not only `deployment.kubernetes.resources.limit` but `deployment.kubernetes.resources.request` as well

----
 # ⚠️ Could not find ticket corresponding to `fix/TECH-XXX_fix_instance_numbers. Does your branch name follow the correct pattern?

### DIFF
--- a/src/mpyl/steps/deploy/k8s/chart.py
+++ b/src/mpyl/steps/deploy/k8s/chart.py
@@ -454,30 +454,36 @@ class ChartBuilder:  # pylint: disable = too-many-instance-attributes
         return V1SealedSecret(name=self.release_name, secrets=secrets)
 
     @staticmethod
-    def _to_resources(resources: Resources, defaults: ResourceDefaults, target: Target):
+    def _to_resource_requirements(
+        resources: Resources, defaults: ResourceDefaults, target: Target
+    ):
         cpus = (
             resources.limit.cpus
-            if resources and resources.limit and resources.limit.cpus
+            if resources.limit and resources.limit.cpus
             else defaults.cpus
         )
-        cpus_limit = cpus.get_value(target=target) * 1000.0
-        cpus_request = (
-            resources.request.cpus
-            if resources and resources.request and resources.request.cpus
+
+        cpus_limit: float = cpus.get_value(target=target) * 1000.0
+
+        cpus_request: float = (
+            resources.request.cpus.get_value(target=target) * 1000.0
+            if resources.request and resources.request.cpus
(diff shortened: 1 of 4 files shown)


src/mpyl/steps/deploy/k8s/chart.py
```diff
@@ -460,13 +460,29 @@ def to_sealed_secrets(self) -> V1SealedSecret:
 
     @staticmethod
     def _to_resources(resources: Resources, defaults: ResourceDefaults, target: Target):
```
Jorg88, line 462: should we maybe rename this function? it's a bit weird to have a `to_resources` function has an input param `resources`

src/mpyl/steps/deploy/k8s/chart.py
```diff
@@ -460,13 +460,29 @@ def to_sealed_secrets(self) -> V1SealedSecret:
...
-        cpus = resources.cpus if resources and resources.cpus else defaults.cpus
+        cpus = (
+            resources.limit.cpus
+            if resources and resources.limit and resources.limit.cpus
```
Jorg88, line 465: we shouldn't have to check for `if resources` since it's not an optional param

Jorg88 (approved):

tests/steps/deploy/k8s/chart/templates/service/deployment.yaml
```diff
@@ -16,7 +16,7 @@ metadata:
     revision: 2ad3293a7675d08bc037ef0846ef55897f38ec8f
   name: dockertest
 spec:
-  replicas: 1
+  replicas: 3
```
SamTheisens, line 19: This is good proof that it works, but perhaps you could also alter one or two of the limit and request fields in the [test_project.yml](https://github.com/Vandebron/mpyl/pull/197/files#diff-cf71122cee41f3d68ad5de7d1cab808c7230c75c448c16bb728c39ccddcc8da0) to see if override and fallback to default work as expected?

SamTheisens (approved):

tests/steps/deploy/k8s/chart/templates/service/deployment.yaml
koeves: thanks, indeed there was a mistake in the scaling

SamTheisens (approved):

Jorg88 (approved):

PR #203, opened 2023-08-30
feature/TECH 547 remove pr namespace after prod deploy

koeves:
delete pr namespace after successful deployment to prod

### DIFF
--- a/src/mpyl/steps/deploy/k8s/__init__.py
+++ b/src/mpyl/steps/deploy/k8s/__init__.py
@@ -73,6 +73,7 @@ def delete_namespace_if_exists(
         logger.error(
             f"Not deleting namespace {namespace}, because cluster config {rancher_config} is not test"
         )
+        return
 
     config.load_kube_config(context=rancher_config.context)
     logger.info(
--- a/src/mpyl/steps/deploy/k8s/__init__.py
+++ b/src/mpyl/steps/deploy/k8s/__init__.py
@@ -4,6 +4,7 @@ from pathlib import Path
 from typing import Optional
 
 from kubernetes import config, client
+from kubernetes.client import V1NamespaceList
 
 from .helm import write_helm_chart
 from ...deploy.k8s.resources import CustomResourceDefinition
@@ -47,7 +48,9 @@ def upsert_namespace(
     api = client.CoreV1Api()
 
     meta_data = rancher_namespace_metadata(namespace, rancher_config)
-    namespaces = api.list_namespace(field_selector=f"metadata.name={namespace}")
+    namespaces: V1NamespaceList = api.list_namespace(
+        field_selector=f"metadata.name={namespace}"
+    )
 
     if len(namespaces.items) == 0 and not dry_run:
         api.create_namespace(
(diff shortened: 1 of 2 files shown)


src/mpyl/steps/deploy/k8s/__init__.py
```diff
@@ -57,6 +60,22 @@ def upsert_namespace(
         logger.info(f"Found namespace {namespace}")
 
 
+def delete_namespace_if_exists(
+    logger: Logger, namespace: str, rancher_config: ClusterConfig
+):
```
Jorg88, line 65: I think we should build in some safety features for this function to only delete for the test cluster and only namespaces that include `pr`

src/mpyl/steps/deploy/k8s/__init__.py
```diff
@@ -90,6 +109,15 @@ def deploy_helm_chart(
         delete_existing,
     )
 
+    if output.success and run_properties.target:
```
Jorg88, line 112: I assume you meant to check for target `== Target.PRODUCTION`?

Jorg88 (changes requested):

src/mpyl/steps/deploy/k8s/__init__.py
koeves: ah yes

src/mpyl/steps/deploy/k8s/__init__.py
koeves: It does only delete in the test cluster

koeves: I'll add a check for namespaces named `pr-XXXX`

src/mpyl/steps/models.py
```diff
@@ -30,6 +30,16 @@ def validate(self) -> Optional[str]:
     def identifier(self) -> str:
         return self.tag if self.tag else f"pr-{self.pr_number}"
 
+    def get_pr_str(self) -> str:
+        pr_num: Optional[str] = (
+            str(self.pr_number)
+            if self.pr_number
+            else (self.tag.split("-")[1] if self.tag else None)
```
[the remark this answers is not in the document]
koeves: no, but you need to be able to figure out what the pr namespace was when you are deploying a tag. a tag is like `20230830-1234` where `1234` is the PR number number, the first part is the date

src/mpyl/steps/deploy/k8s/__init__.py
```diff
@@ -57,6 +60,22 @@ def upsert_namespace(
...
+    config.load_kube_config(context=rancher_config.context)
+    logger.info(
+        f"Deleting namespace {namespace} from k8s context {rancher_config.context} "
+        f"after successful deployment to production."
+    )
+    api = client.CoreV1Api()
+    namespaces: V1NamespaceList = api.list_namespace(
+        field_selector=f"metadata.name={namespace}"
+    )
+    if len(namespaces.items) == 1:
```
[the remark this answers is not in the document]
koeves: Namespace names are unique. So we would be reimplementing kubernetes' checks in this case

src/mpyl/steps/deploy/k8s/__init__.py
Jorg88: I know you pass the test cluster config here, but it's completely possible to pass the config for production right?

src/mpyl/steps/deploy/k8s/__init__.py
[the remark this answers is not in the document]
Jorg88: Ah fair, I got thrown off because of the `== 1` instead of `> 0`

src/mpyl/steps/deploy/k8s/__init__.py
koeves: ah you mean the method itself

koeves: it's not easy since there's no guarantees what the test clusters name is.. And in the end it's just a ClusterConfig

Jorg88: You can check for `clusterEnv`?

koeves: fair

Jorg88 (approved):

koeves:
Closing in favor of automated job

PR #185, opened 2023-08-15
feature/TECH 507 argo cd support

SamTheisens:
Resolves #181

- [x] Introduce plain k8s manifests as deploy result. These can be deployed with `kubectly apply -f`, or feed into argocd.
- [x] Replace brittle dictionary fiddling
- [x] Move type definitions local to their respective domains
- [x] Clearly describe the 4 different targets of a k8s deploy in documentation 

⚠️ `mpyl_config.yml` will need to have the imagePullSecrets defined like such:
```yaml
    kubernetes:
      imagePullSecrets:
        - name: 'acme-registry'
```

----
### 📕 [TECH-507](https://vandebron.atlassian.net/browse/TECH-507) Integrate MPyL with ArgoCD sam@vandebron.nl 
[181](https://github.com/Vandebron/mpyl/issues/181) 

🏗️ Build [17](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-185/17/display/redirect) ✅ Successful, started by _Sam Theisens_  
🚀 *[cloudfront-service](https://cloudfront-service-185.test.nl/)*, *job*, *kong-sync*, *[nodeservice](https://nodeservice-185.test.nl/)*, *[sbtservice](https://sbtservice-185.test.nl/)*, *sparkJob*  

[TECH-507]: https://vandebron.atlassian.net/browse/TECH-507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #212, opened 2023-09-05
[TECH-117] Support multiple Docker registries

peterrutg:
----
### 📕 [TECH-117](https://vandebron.atlassian.net/browse/TECH-117) Specify external registry per module peterrutgers@vandebron.nl 
In order to be able to deploy to different registries based on the projects that we are building.

Needed for NUC deployments.

**How**

* Check how it’s done in MPL
* Improve upon it in MPyL → `Docker After Build` step
* Default config should be bigdataregistry, but we should be able to override that with a config

🏗️ Build [10](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-212/10/display/redirect) ✅ Successful, started by _Sam Theisens_  
🚀 *[cloudfront-service](https://cloudfront-service-212.test.nl/swagger/index.html)*, *example-dagster-user-code*, *job*, *kong-sync*, *[nodeservice](https://nodeservice-212.test.nl/swagger/index.html)*, *[sbtservice](https://sbtservice-212.test.nl/swagger/index.html)*, *sparkJob*  

[TECH-117]: https://vandebron.atlassian.net/browse/TECH-117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/mpyl/utilities/docker/__init__.py
+++ b/src/mpyl/utilities/docker/__init__.py
@@ -174,7 +174,7 @@ def push_to_registry(
     image = docker.image.inspect(image_name)
     logger.debug(f"Found image {image}")
 
-    login(logger=logger, docker_config=docker_config)
+    login(logger=logger, registry_config=docker_config)
     full_image_path = docker_registry_path(docker_config, image_name)
     docker.image.tag(image, full_image_path)
     docker.image.push(full_image_path, quiet=False)
@@ -288,14 +288,14 @@ def build(
         return False
 
 
-def login(logger: Logger, docker_config: DockerRegistryConfig) -> None:
-    logger.info(f"Logging in with user '{docker_config.user_name}'")
+def login(logger: Logger, registry_config: DockerRegistryConfig) -> None:
+    logger.info(f"Logging in with user '{registry_config.user_name}'")
     docker.login(
-        server=f"https://{docker_config.host_name}",
-        username=docker_config.user_name,
-        password=docker_config.password,
+        server=f"https://{registry_config.host_name}",
+        username=registry_config.user_name,
+        password=registry_config.password,
     )
-    logger.debug(f"Logged in as '{docker_config.user_name}'")
+    logger.debug(f"Logged in as '{registry_config.user_name}'")
 
 
 def create_container(logger: Logger, image_name: str) -> Container:
--- a/src/mpyl/utilities/docker/__init__.py
+++ b/src/mpyl/utilities/docker/__init__.py
@@ -97,7 +97,7 @@ class DockerConfig:
             registries: dict = config["docker"]["registries"]
             build_config: dict = config["docker"]["build"]
             return DockerConfig(
-                default_registry=config["docker"]["default_registry"],
+                default_registry=config["docker"]["defaultRegistry"],
                 registries=[DockerRegistryConfig.from_dict(r) for r in registries],
                 root_folder=build_config["rootFolder"],
                 build_target=build_config.get("buildTarget", None),
--- a/src/mpyl/utilities/docker/__init__.py
+++ b/src/mpyl/utilities/docker/__init__.py
@@ -183,13 +183,13 @@ def push_to_registry(
 def registry_for_project(
     docker_config: DockerConfig, project: Project
 ) -> DockerRegistryConfig:
-    if project.docker:
-        host_name = project.docker.host_name
-    else:
-        host_name = docker_config.default_registry
-    for registry in docker_config.registries:
-        if registry.host_name == host_name:
-            return registry
+    host_name = (
+        project.docker.host_name if project.docker else docker_config.default_registry
+    )
+    registry = next(r for r in docker_config.registries if r.host_name == host_name)
+    if registry:
+        return registry
+
     raise KeyError(f"Docker config has no registry with host name {host_name}")
 
 
--- a/src/mpyl/utilities/docker/__init__.py
+++ b/src/mpyl/utilities/docker/__init__.py
@@ -233,7 +233,7 @@ def build(
     file_path: str,
     image_tag: str,
     target: str,
-    docker_config: Optional[DockerRegistryConfig] = None,
+    registry_config: Optional[DockerRegistryConfig] = None,
 ) -> bool:
     """
     :param logger: the logger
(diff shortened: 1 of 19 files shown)


SamTheisens:
@peterrutg I think the registry should be configurable per `project.yml`? 
It would fit snugly next to https://github.com/Vandebron/mpyl/pull/185/files#diff-e55c3e2776c6472317b225ad77d25d58a822c4642672265de0de77e437a69073R779-R787 I believe.

src/mpyl/utilities/docker/__init__.py
```diff
@@ -144,6 +158,16 @@ def docker_registry_path(docker_config: DockerConfig, image_name: str) -> str:
...
 
 
+def registry_for_project(
+    docker_config: DockerConfig, project_docker: Optional[Docker]
+) -> DockerRegistryConfig:
+    if project_docker:
+        for registry in docker_config.registries:
+            if registry.host_name == project_docker.registry_url:
+                return registry
+    return docker_config.registries[0]
```
Jorg88, line 168: Should we allow for a fallback or should be actually throw an error here (you know better than me)?

Jorg88 (commented):
I assume you'll add a unit test later in which there actually are multiple registries configured/available and you want to pick a specific one?

peterrutg:
> @peterrutg I think the registry should be configurable per `project.yml`? It would fit snugly next to https://github.com/Vandebron/mpyl/pull/185/files#diff-e55c3e2776c6472317b225ad77d25d58a822c4642672265de0de77e437a69073R779-R787 I believe.

Yes the idea was to reuse the `docker.registryUrl` which was already defined in `project.yml` https://github.com/Vandebron/mpyl/blob/main/src/mpyl/schema/project.schema.yml#L103

There's also `docker.credentialsId` in there, which was used by MPL, but we were thinking of replacing that one by username + password from the config file. So that the `project.yml` points to one of the entries in `mpyl_config.yml`

src/mpyl/utilities/docker/__init__.py
peterrutg: Most `project.yml` have no `docker` block so think we still need default behavior. Either taking the first one or having one marked as default explicitly?

SamTheisens: Why don't we define custom docker registries in `project.yml` only? That seems more in line with the way we've been using `mpyl_config.yml` so far: to provide global settings and `project.yml` defaults.

koeves: Also, currently the `docker` field in the project yaml has the following structure:
```
docker:
  credentialsId: 'dockerExternalRegistry'
  registryUrl: 'vdbexternalaccessregistry.azurecr.io'
```
I think we need to change the schema and the mechanism here, because we're referring to a jenkins credential here, which I don't think we have access to in mpyl

SamTheisens: Right, I think I understand the problem better now.
The key thing I misunderstood is that this is configuration that is needed during the _build_ stage. I see two ways to solve this:
1. _Turn de docker registry config into a list as @peterrutg proposed and reference it from `project.yml`_
   - *Pro* avoids credential config duplication of `project.yml`
   - *Pro* allows us to inject build time credentials in a way that we're already used to.
   - *Con* introduces reference from `project.yml` to `mpyl_conig.yml` that is not immediately obvious to the reader and cannot be validated by the schema.
   - *Con* introduces breaking change to the schema. 
2. _Define a single default docker registry in the `mpyl_config.yml` and allow this to be overridden at `project.yml` level._
   - *Pro* quite explicit and which projects will get non-default values. Especially in scenarios where we have many (say > 10) registries used by only specific projects, defining them in the `project.yml` would be explicit and reduce the chances of stale configuration in `mpyl_config.yml` that nobody knows whether it's still in use.
   - *Con* no possibility to do build time env substitution of `project.yml`. This shouldn't be a difficult feature to add, but it makes the project config less declarative. It may not be obvious to the user when the variables will be substituted. And perhaps also open the door to "powerful" or "clever" possibilities that I can't foresee at the moment (project.yml override comes to mind).  

I'm personally tending towards option 1, with a few implementation notes:
 - would benefit from automated upgrade script for `mpyl_config.yml` as introduced by this PR https://github.com/Vandebron/mpyl/pull/253/commits/0612f26b1c5881e7b74b48ac2cea501ab6f44a3c
 - I would suggest to introduce a default registry field to avoid having to define the registry to use in all `project.yml`. Something like this:
 ```
 docker:
  defaultRegistry: 'bigdataregistry.azurecr.io'
  registries:
    - hostName: 'bigdataregistry.azurecr.io'
      userName: !ENV ${DOCKER_REGISTRY_USR:docker_user}
      password: !ENV ${DOCKER_REGISTRY_PSW:docker_password}
    - hostName: 'vdbexternalaccessregistry.azurecr.io'
      userName: !ENV ${DOCKER_EXTERNAL_REGISTRY_USR:docker_user}
      password: !ENV ${DOCKER_EXTERNAL_REGISTRY_PSW:docker_password}
 ```
  - The `credentialsId` and `registryUrl` fields aren't used by MPyL so I think we can deprecate and remove them.
  - What about the `imagePullSecrets` that you need deploy time? I guess they relate to the registry that you've chosen to push your images too. Although that doesn't _have_ to be the case, in case we want to use images built by an external party.
  
@koeves @peterrutg What do you think?

SamTheisens: https://github.com/Vandebron/mpyl/pull/253 with config upgrade mechanism is ready for review.

SamTheisens (approved):
Looks good 👍 . 
I've taken the liberty to rebase onto https://github.com/Vandebron/mpyl/pull/253, which now also tests the upgrade procedure as part of the integration tests: https://github.com/Vandebron/mpyl/pull/253/commits/183d8a98d8c8109b42dcf40cd54299ebcd75cc9b#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R45

src/mpyl/utilities/docker/__init__.py
SamTheisens: @peterrutg the config upgrade is in master. I've rebased this branch.

peterrutg: Cool, thanks!

koeves (approved):

PR #218, opened 2023-09-05
Feature/tech 521 implement remote artifact caching

Jorg88:


src/mpyl/cli/artifacts.py
```diff
@@ -48,10 +52,28 @@ def pull(obj: CliContext, tag, pr):
...
+    artifact_repo_path = Path("../" + artifact_repo_config.folder)
+
+    try:
+        clone_repository(artifact_repo_config, artifact_repo_path)
+        artifact_repo = Repository(
+            config=artifact_repo_config, root_dir=artifact_repo_path
+        )
+        build_artifacts = BuildArtifacts(
+            logger=logger,
+            codebase_repo=obj.repo,
```
SamTheisens: Why does this need to be a repo? Isn't the local filesystem enough?

src/mpyl/artifacts/__init__.py
```diff
@@ -6,23 +6,26 @@
...
         codebase_repo: Repository,
         artifact_repo: Repository,
         artifact_repo_path: Path,
-    ):  # codebase repo might not be needed
+    ):
         self.logger = logger
         self.codebase_repo = codebase_repo
         self.artifact_repo = artifact_repo
-        self.artifact_repo_path = artifact_repo_path
+        self.cache_folder = artifact_repo_path / CACHE_FOLDER_NAME
```
SamTheisens: Can't we use a temp folder for this?

src/mpyl/utilities/repo/__init__.py
```diff
@@ -271,8 +274,9 @@ def checkout_branch(self, branch_name: str):
         self._repo.git.switch(branch_name)
 
     def does_branch_exist(self, branch_name: str, remote=False) -> bool:
+        self._repo.remote().fetch()
         found_branches = [
-            branch.strip(" ")
+            branch.strip(" ").strip("*")
```
SamTheisens, line 279: What is this * for?

src/mpyl/artifacts/__init__.py
```diff
@@ -40,11 +40,17 @@ def get_build_artifacts_paths(self) -> list[Path]:
...
+            self.logger.info(f"Fetching branch '{branch}' from remote")
+            self.artifact_repo.fetch_branch(branch_name=branch)
+            self.artifact_repo.checkout_branch(branch_name=branch)
+
+            for artifact_path in os.listdir(self.cache_folder):
+                shutil.copytree(
+                    src=self.cache_folder / artifact_path,
+                    dst=artifact_path,
+                    dirs_exist_ok=True,
+                )
```
SamTheisens: I think it would be good to make a clear distinction between the gathering and copying of artifacts (from the filesystem) and the repo operations. In principle you could do everything without every using a _remote_ repo. Including unit testing.

src/mpyl/cli/artifacts.py
```diff
@@ -39,11 +42,11 @@ def artifacts(ctx: click.Context, config: str, verbose: bool):
 
 
 @artifacts.command(help="Pull artifacts")
-@click.option("--tag", "-t", help="Tag to build", type=click.STRING, required=False)
-@click.option("--pr", type=click.INT, help="PR number to fetch", required=False)
+@click.option("--tag", "-t", type=click.STRING, help="Tag to build", required=False)
+@click.option("--pr_number", type=click.INT, help="PR number to fetch", required=False)
```
SamTheisens, line 58: what about `--pr_number` and `-pr` as short?

src/mpyl/cli/build.py
```diff
@@ -84,8 +85,15 @@ async def warn_if_update(console: Console):
     show_default=True,
     help="Verbose output",
 )
+@click.option(
+    "--cached",
+    help="Use cached build artifacts",
+    is_flag=True,
+    required=False,
+    default=False,
+)
```
SamTheisens, line 94: I think I would find it cleaner to leave the caching outside the build step. It's optional anyway, so thing will still work if you forget to d it.

SamTheisens (changes requested):
Looks good in general 👍 .
I guess we need the following operations:
 - find artifacts to cache
 - create or clone a repo in a temp folder 
 - copy the artifacts to the folder
 - push the changes to the artifact repo

And for pull
 - clone the artifacts repo to a temp folder
 - copy the artifacts onto your local FS

I think these operations can be kept mostly independent. You could start with filesystem copy only. And then link up the git persistence operations later.

src/mpyl/artifacts/__init__.py
Jorg88: What do you mean precisely? I needed a subfolder inside the repo for handling all the build artifacts logic (like deleting everything before a copy) and avoid issues with the existing files that you get from cloning (like the readme and .gitignore).

src/mpyl/utilities/repo/__init__.py
Jorg88: To remove the `*` that is placed before the current active branch when you do `git branch --list`

src/mpyl/cli/build.py
Jorg88: Actually @koeves proposed to put it in build since it can be used by all build commands

src/mpyl/utilities/github/__init__.py
```diff
@@ -58,3 +63,11 @@ def get_pr_for_branch(repo: Repository, branch: str) -> PullRequest:
...
 
     return pulls.pop()
+
+
+def clone_repository(artifact_repo_config: RepoConfig, repo_path: Path) -> None:
+    repo_path.mkdir(parents=True, exist_ok=True)
+    custom_check_output(
+        logging.getLogger(),
+        f"git clone {artifact_repo_config.repo_credentials.ssh_url} {repo_path}",
+    )
```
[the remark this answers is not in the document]
Jorg88: Hmm, we need to align on how the function should look though. I kinda prefer this approach where you just pass the config (or config path) and all the details required for the clone are taken from there. Also, I don't think branch should be mandatory? And lastly, your pr is only a draft pr.. I can't implement it (except copy pasting, which might become a mess) yet 😬 

Does it make sense to say that if my pr gets merged first you can adjust this as part of your pr and the other way around?

src/mpyl/cli/artifacts.py
```diff
@@ -48,10 +52,28 @@ def pull(obj: CliContext, tag, pr):
...
+        build_artifacts = BuildArtifacts(
+            logger=logger,
+            codebase_repo=obj.repo,
+            artifact_repo=artifact_repo,
+            artifact_repo_path=artifact_repo_path,
+        )
+        build_artifacts.push(branch=target_branch)
+    finally:
+        print()
+        # shutil.rmtree(artifact_repo_path, ignore_errors=True)
```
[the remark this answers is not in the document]
Jorg88: Ah you were reviewing an old commit, I was wondering how I missed this 😅

src/mpyl/cli/artifacts.py
Jorg88: I don't see how unless you're proposing a completely different solution? A lot of the logic in `BuildArtifacts` uses the repo and I think it looks pretty clean actually?

Jorg88:
Adressed the comments that I could @SamTheisens, maybe we can discuss the rest on slack. Did you know you were looking at an old version for some of the comments?

Also, do you have any idea where the test error comes from since it works fine locally?

src/mpyl/utilities/github/__init__.py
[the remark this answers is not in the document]
SamTheisens: Use `Repo.clone_from(` at least. `custom_check_output(` should be avoided. See warning in its documentation.

src/mpyl/artifacts/__init__.py
SamTheisens: A temp folder is located outside the repo path https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

Jorg88: It's not a temp folder, how am I supposed to put the folder which contains the files that need to be pushed/pulled from/in outside the repo?

SamTheisens: create temp folder, clone or create repo in it, copy files, commit, push

Jorg88: Ahh, I understood it differently because you commented on a folder that is _inside_ the repo folder. I'll take a look at this TemDir python thing, didn't know about it!

src/mpyl/artifacts/__init__.py
Jorg88: In theory - yes. But in practice 90% of this feature is interacting with a remote repo. If you would split it up and you would want to unit test without using a remote repo, you would basically just be testing the copying of the files from the repo to the build folder and the other way around. Which is 1 line of code, so is that worth the extra complexity unless we are sure we want to use different sources than a repo?

src/mpyl/cli/artifacts.py
```diff
@@ -0,0 +1,105 @@
...
+import click
+
+from ..artifacts.build_artifacts import BuildArtifacts
+from ..cli import CliContext, create_console_logger, CONFIG_PATH_HELP
+from ..constants import DEFAULT_CONFIG_FILE_NAME, DEFAULT_RUN_PROPERTIES_FILE_NAME
+from ..steps.models import RunProperties
+from ..utilities.repo import Repository, RepoConfig
+
+
+@click.group("artifacts")
```
SamTheisens, line 16: I would expect `push` and `pop` to be filed under `build` intuitively.
The workflow would be
```
mpyl build pop
mpyl build run
mpyl build push
```
WDYT?

Jorg88: Well this is how we refined it a few weeks ago during our call, we decided to implement it in 2 ways:
- A toggle to the build commands
- Its own cli commands

I did look into creating an 'artifacts' subgroup in `build` but that didn't seem possible.

SamTheisens: On second thought I prefer the way I suggested. Don't you?

Jorg88: Eh it's just semantics really, programmatically I prefer `mpyl artifacts pull/push` since `build` is already big enough by itself. It's also more explicit to what you're actually doing compared to `build pop`. But it's not a big deal for me, you can decide with @koeves next week since he'll take over the pr while I'm on vacation.

src/mpyl/cli/build.py
```diff
@@ -145,9 +156,18 @@ def run(obj: CliContext, ci, all_, tag):  # pylint: disable=invalid-name
...
+
+    with TemporaryDirectory(dir=".", prefix=".artifacts-") as tmp_dir:
+        build_artifacts = _prepare_artifacts_repo(obj=obj, repo_path=Path(tmp_dir))
+        if obj.cached and branch:
+            build_artifacts.pull(branch=branch)
+        result = run_mpyl(
+            run_properties=run_properties, cli_parameters=parameters, reporter=None
+        )
+        if branch:
+            build_artifacts.push(branch=branch)
```
SamTheisens, line 170: I would really take this logic out of the build step

src/mpyl/cli/artifacts.py
```diff
@@ -0,0 +1,105 @@
...
+    run_properties = RunProperties.from_configuration(obj.run_properties, obj.config)
+    target_branch = (
+        tag if tag else f"PR-{pr_number or run_properties.versioning.pr_number}"
+    )
+    if not target_branch:
+        raise click.ClickException("Either --pr or --tag must be specified")
+
+    with TemporaryDirectory(dir=".", prefix=".artifacts-") as tmp_dir:
+        build_artifacts = _prepare_artifacts_repo(obj=obj, repo_path=Path(tmp_dir))
+        build_artifacts.pull(branch=target_branch)
```
SamTheisens, line 69: I think it would be cleaner to pull the temporary directory inside the BuildArtifacts class

PR #220, opened 2023-09-06
Run validation tasks in parallel

### LINKED ISSUE #219: As a developer I want a quick validation roundtrip
The `pipenv run validate` script runs the following checks:
```
pipenv run format
pipenv run lint
pipenv run lint-test
pipenv run check-types
pipenv run check-types-test
pipenv run test
```
This takes 40 seconds at the time of writing and will only take more time going forward.

These tasks could all easily run in parallel, as they do in our github actions. A shorter feedback loop would significantly improve the developer experience.

SamTheisens:
Resolves #219
Replaces the combined validation script:
```toml
validate = "bash -c 'black --check --diff . && mypy --explicit-package-bases src/mpyl/ && mypy --explicit-package-bases --check-untyped-defs tests/ && pytest -W ignore::pytest.PytestCollectionWarning && pylint src/ && pylint --rcfile .pylintrc-test tests/'"
```
 with a script that executes the tasks in parallel.

validate

----
 # ⚠️ Could not find ticket corresponding to `feature/quick-dev-feedback-loop. Does your branch name follow the correct pattern?


Jorg88 (approved):
Very fancy Sam!

koeves (approved):
nice!

PR #226, opened 2023-09-06
feature/TECH 565 Remove implicit deploy logic

koeves:
- Remove implicit deploy logic after PROD deployment
- Attempt to fix swagger links

[TECH-565]: https://vandebron.atlassian.net/browse/TECH-565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TECH-565]: https://vandebron.atlassian.net/browse/TECH-565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

----
### 📕 [TECH-565](https://vandebron.atlassian.net/browse/TECH-565) Remove logic to deploy to test after deploy to prod danielkoves@vandebron.nl 

🏗️ Build [3](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-226/3/display/redirect) ✅ Successful, started by _Daniel Koves_  
🚀 *[cloudfront-service](https://cloudfront-service-226.test.nl/swagger/index.html)*, *example-dagster-user-code*, *job*, *kong-sync*, *[nodeservice](https://nodeservice-226.test.nl/swagger/index.html)*, *[sbtservice](https://sbtservice-226.test.nl/swagger/index.html)*, *sparkJob*

### DIFF
--- a/src/mpyl/steps/deploy/kubernetes.py
+++ b/src/mpyl/steps/deploy/kubernetes.py
@@ -90,7 +90,7 @@ class DeployKubernetes(Step):
                     self._logger,
                     chart,
                     step_input,
-                    Target.PRODUCTION,
+                    Target.PULL_REQUEST_BASE,
                     builder.release_name,
                 )
                 return Output(
--- a/src/mpyl/steps/deploy/kubernetes.py
+++ b/src/mpyl/steps/deploy/kubernetes.py
@@ -81,26 +81,6 @@ class DeployKubernetes(Step):
                 step_input,
                 spec=DeployedHelmAppSpec(url=f"{url}"),
             )
-            if properties.target == Target.PRODUCTION:
-                self._logger.info(
-                    f"Release to production successful, updating base images in {Target.PULL_REQUEST_BASE} "
-                    f"to make sure the Test environment is in sync with production"
-                )
-                deploy_to_prod_result = deploy_helm_chart(
-                    self._logger,
-                    chart,
-                    step_input,
-                    Target.PULL_REQUEST_BASE,
-                    builder.release_name,
-                )
-                return Output(
-                    success=True,
-                    message=deploy_result.message
-                    + "\n"
-                    + deploy_to_prod_result.message,
-                    produced_artifact=artifact,
-                )
-
             return Output(
                 success=True, message=deploy_result.message, produced_artifact=artifact
             )
--- a/src/mpyl/steps/deploy/kubernetes.py
+++ b/src/mpyl/steps/deploy/kubernetes.py
@@ -67,11 +67,11 @@ class DeployKubernetes(Step):
             if hostname:
                 has_specific_routes_configured: bool = bool(
                     builder.deployment.traefik is not None
+                    and properties.target == Target.PRODUCTION
                 )
                 self._logger.info(
                     f"Service {step_input.project.name} reachable at: {hostname}"
                 )
-
                 endpoint = (
                     "/" if has_specific_routes_configured else "/swagger/index.html"
                 )
--- a/src/mpyl/steps/steps.py
+++ b/src/mpyl/steps/steps.py
@@ -67,6 +67,7 @@ class Steps:
         artifact: Optional[Artifact],
         dry_run: bool = False,
     ) -> Output:
+        self._logger.info(f"Executing {executor.meta.name} for '{project.name}'")
         required = executor.required_artifact
         if (
             artifact is not None


src/mpyl/steps/deploy/kubernetes.py
```diff
@@ -85,7 +85,7 @@ def execute(self, step_input: Input) -> Output:
                     self._logger,
                     chart,
                     step_input,
-                    Target.PRODUCTION,
+                    Target.PULL_REQUEST_BASE,
```
SamTheisens, line 88: Wow! We were deploying to prod twice? :-)
Maybe it would be cleaner to do this in the runner instead of the step?

src/mpyl/build.py
```diff
@@ -33,13 +33,13 @@ def get_build_plan(
     branch = repo.get_branch or run_properties.versioning.branch
     if branch:
         if repo.base_revision:
-            logger.info(
+            logger.debug(
                 f"Branch `{repo.main_branch}` already present locally. Skipping pull."
             )
         else:
-            logger.info(f"Pulling `{repo.main_branch}` from {repo.remote_url}")
+            logger.debug(f"Pulling `{repo.main_branch}` from {repo.remote_url}")
```
SamTheisens, line 40: This shouldn't be necessary anymore I think? Now that we have `mpyl repo init`.

src/mpyl/steps/deploy/kubernetes.py
koeves: Yeah maybe we can remove this logic from mpyl.. It's a bit implicit, and with the proper caching mechanism in place it shouldn't be needed either

SamTheisens: Agreed. It's too implicit 👍 
Why would a proper caching mechanism make this logic redundant?

koeves: I thought we're doing this to ensure the same image gets deployed to test which is in prod. This already was the case when the work is scheduled on the same agent. With the artifact repo, it will always be the case right?

SamTheisens: Ah, I understand what you mean by caching now. I'm not sure about that. Because although the images used in test should be identical to production, the config will not be.

koeves: wait.. currently if you're scheduled on the same agent and you're deploying a tag it's not rebuilding the image

src/mpyl/steps/deploy/kubernetes.py
```diff
@@ -64,11 +64,11 @@ def execute(self, step_input: Input) -> Output:
...
                     builder.deployment.traefik is not None
+                    and properties.target == Target.PRODUCTION
                 )
                 self._logger.info(
                     f"Service {step_input.project.name} reachable at: {hostname}"
                 )
-
                 endpoint = (
                     "/" if has_specific_routes_configured else "/swagger/index.html"
                 )
```
koeves, line 74: @SamTheisens I have a question about this logic as well. So currently if you have a traefik config in your service we don't attach `/swagger/index.html` to the URL which then gets put in the PR descriptions. This means that for most of our services, the link in PR description is broken. 

Just because you have traefik host overrides, the swagger documentation is going to be in the same place though right? I mean, the host can be specific for the service but the path of the swagger doc is not?

src/mpyl/build.py
koeves: Yeah, but this method is used in the mpyl build status as well. Which is a cli command, maybe best to leave it in?

src/mpyl/steps/deploy/kubernetes.py
SamTheisens: Yeah, I think so. The default rules are always applied AFAIK. You could check in traefik if that's indeed the case. We could  re-introduce `/swagger/index.html`. Although it's very Vandebron specific...

src/mpyl/build.py
SamTheisens: I think it would be better to error out in that case end instruct people what to do.

Jorg88 (approved):

src/mpyl/steps/deploy/kubernetes.py
koeves: Do you have a better suggestion to always be able to figure out the swagger endpoint? Should we provide an option al swagger host in the project yaml?

PR #236, opened 2023-09-08
[TECH-435] Add dryrun and version flag

ewaldschrader:
branch: feature/TECH-435-IMPROVING-CLI

----
### 📕 [TECH-435](https://vandebron.atlassian.net/browse/TECH-435) CLI improvements ewaldschrader@vandebron.nl 
Add support for the following things in the CLI

* do a dry run
* set mpyl version of jenkins run (currently we can only set test version)
* deploy parameters 
* other build parameters?

----

We should be able to set the parameters for Jenkins via the cli like:

```
mpyl build jenkins --argument "run run --dryrun"```


→ this probably doesn’t work with `click` . 

- Define custom argument type with click
- Parse string argument

Another (maybe better) option is to just add a dryrun flag for mpyl cli, and then pass that on to jenkins via the jenkins API:

{noformat}mpyl build jenkins --dryrun
```
```

🏗️ Build [2](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-236/2/display/redirect) ✅ Successful, started by _Ewald Schrader_  
🚀 *[cloudfront-service](https://cloudfront-service-236.test.nl/)*, *example-dagster-user-code*, *job*, *kong-sync*, *[nodeservice](https://nodeservice-236.test.nl/)*, *[sbtservice](https://sbtservice-236.test.nl/)*, *sparkJob*  


[TECH-435]: https://vandebron.atlassian.net/browse/TECH-435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### CODE REVIEW

File: Jenkinsfile [hunk 1]
```diff
@@ -39,7 +40,7 @@ pipeline {
                     withKubeConfig([credentialsId: 'jenkins-rancher-service-account-kubeconfig-test']) {
                         wrap([$class: 'BuildUser']) {
                             sh "pipenv clean"
-                            sh "pipenv install --ignore-pipfile --skip-lock --site-packages --index https://test.pypi.org/simple/ 'mpyl==$CHANGE_ID.*'"
+                            sh "pipenv install --ignore-pipfile --skip-lock --site-packages --index 'mpyl==${params.MPYL_RELEASE}'"
```
Jorg88, line 43: Did you check what happens when you pass a normal version of mpyl instead of a test version?

src/mpyl/cli/build.py
```diff
@@ -262,11 +272,34 @@ def by_choice(choice: Choice):
...
+    versions.sort(key=LooseVersion, reverse=True)
+    versions = [version[:-4] for version in versions[:100]]
+    versions = list(dict.fromkeys(versions))
+    versions = [version + "*" for version in versions]
+    return versions
+
+
+def select_version() -> str:
+    console = Console()
+    console.log("Fetching MPyL releases..")
```
Jorg88, line 288: I would put this log in the function above (since that actually fetches the releases)

Jenkinsfile
ewaldschrader: Good point. I've added a version check in the `select_version` function. It is only possible to install test version with the `--version` flag. You think it's necessary to be able to install normal versions as well?

PR #251, opened 2023-09-19
Readd Dagster definition section in mpyl-config-yaml-schema

kgrunert:
@SamTheisens  somehow the release notes commit 1de6bc5a0fae3e8f65586712e2bc7905a8d26f35 got rid of the Dagster definition section in the schema.
If that was not an accident, let's discuss :-) if it was an accident, this is the PR to revert it 😄 

----
 # ⚠️ Could not find ticket corresponding to `fix/readd-dagster-definition-in-yaml. Does your branch name follow the correct pattern?

### DIFF
--- a/src/mpyl/schema/mpyl_config.schema.yml
+++ b/src/mpyl/schema/mpyl_config.schema.yml
@@ -384,3 +384,23 @@ definitions:
     dependencies:
       accessKeyId: [ secretAccessKey ]
       secretAccessKey: [ accessKeyId ]
+  Dagster:
+    title: Dagster
+    type: object
+    additionalProperties: false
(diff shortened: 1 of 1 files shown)


SamTheisens (approved):
Definitely a mistake! 😱  I would never want to be so rude :-)
@kgrunert would you like us to make a 1.2.1 release with this fix?

kgrunert:
> Definitely a mistake! 😱 I would never want to be so rude :-) 
I also thought so. That doesnt match your character!

> @kgrunert would you like us to make a 1.2.1 release with this fix?
That would be great 🙏  I will run the pipeline test and merge

PR #254, opened 2023-09-21
[TECH-435] add dryrun, version, all flag

ewaldschrader:
branch: feature/TECH-435-CLI_IMPROVEMENTS

----
### 📕 [TECH-435](https://vandebron.atlassian.net/browse/TECH-435) CLI improvements ewaldschrader@vandebron.nl 
Add support for the following things in the CLI

* do a dry run
* set mpyl version of jenkins run (currently we can only set test version)
* deploy parameters 
* other build parameters?

----

We should be able to set the parameters for Jenkins via the cli like:

```
mpyl build jenkins --argument "run run --dryrun"```


→ this probably doesn’t work with `click` . 

- Define custom argument type with click
- Parse string argument

Another (maybe better) option is to just add a dryrun flag for mpyl cli, and then pass that on to jenkins via the jenkins API:

{noformat}mpyl build jenkins --dryrun
```
```

🏗️ Build [75](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-254/75/display/redirect) 🦥 Nothing to do, started by _Ewald Schrader_  


[TECH-435]: https://vandebron.atlassian.net/browse/TECH-435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### CODE REVIEW

File: src/mpyl/cli/build.py [hunk 1]
```diff
@@ -360,10 +404,34 @@ def ask_for_tag_input(ctx, _param, value) -> Optional[str]:
...
-    callback=ask_for_tag_input,
+    callback=ask_for_input,
+)
+@click.option(
+    "--version",
+    "-v",
+    is_flag=False,
+    flag_value="prompt",
+    default="not_set",
+    envvar="MPYL_RELEASE",
```
SamTheisens, line 415: I don't understand in what scenario you would want to set this version via an env var. the jenkins command never runs in a CI context.

src/mpyl/schema/run_properties.schema.yml
```diff
@@ -91,6 +91,9 @@ properties:
           tag:
             description: reference that points to the MPyL version
             type: ["string", "null"]
+          version:
+            description: MPyL version to be installed
+            type: [ "string", "null" ]
```
SamTheisens, line 96: These are CI parameters, so I think we don't need this?

src/mpyl/cli/build.py
```diff
@@ -281,11 +323,13 @@ def select_target():
...
 
 
-def ask_for_tag_input(ctx, _param, value) -> Optional[str]:
+def ask_for_input(ctx, _param, value) -> Optional[str]:
     if value == "not_set":
         return None
-    if value == "prompt":
+    if value == "prompt" and str(_param) == "<Option tag>":
         return select_tag(ctx)
+    if str(_param) == "<Option version>":
```
SamTheisens, line 331: These strings ("<Option version>", "<Option tag>") are probably best defined as constants.

src/mpyl/cli/build.py
```diff
@@ -281,11 +323,13 @@ def select_target():
     ).ask()
 
 
-def ask_for_tag_input(ctx, _param, value) -> Optional[str]:
+def ask_for_input(ctx, _param, value) -> Optional[str]:
     if value == "not_set":
```
SamTheisens, line 327: Maybe we can add `sys.stdout.isatty()` to this condition, to prevent input from ever being asked in CI scenarios.

src/mpyl/cli/build.py
```diff
@@ -270,6 +280,38 @@ def by_choice(choice: Choice):
...
+    console = Console()
+    console.log("Fetching MPyL releases..")
+    url = "https://test.pypi.org/pypi/mpyl/json"
+    data = requests.get(url, timeout=30).json()
+    versions = list(data["releases"].keys())
+    versions.sort(key=LooseVersion, reverse=True)
+    versions = [version[:-4] for version in versions[:100]]
+    versions = list(dict.fromkeys(versions))
+    versions = [version + "*" for version in versions]
+    return versions
```
SamTheisens, line 293: You could unittest this logic by capturing a json response.

src/mpyl/cli/build.py
```diff
@@ -115,15 +117,23 @@ def build(ctx, config, properties, verbose):
     is_flag=True,
     help="Build all projects, regardless of changes on branch",
 )
+@click.option(
+    "--dryrun",
+    "dryrun_",
+    is_flag=True,
+    default=False,
+    help="don't push or deploy images",
```
SamTheisens, line 125: This illustrates why `dryrun` means different things in different contexts. 
I think we should stop passing it along to individual steps, but rather map the CLI flag to what it means locally e.g. don't push image, write manifest instead of deploy, etc.

SamTheisens (changes requested):
Looks good in general. I take it that the primary use case is for testing PR builds on the monorepo? I wonder if that's still a pressing case now that we've started making release candidates?

src/mpyl/cli/build.py
ewaldschrader: added a unittest

src/mpyl/cli/build.py
ewaldschrader: @SamTheisens Yes I see what you mean.. Perhaps better to change in the next PR?

PR #255, opened 2023-09-24
Customizable stages

SamTheisens:
as an intermediate step towards making them more dynamic

branch: feature/TECH-293-flexible-stage-names


src/mpyl/project.py
```diff
@@ -51,14 +51,9 @@ def __str__(self):
...
-    def __eq__(self, other):
-        return self.value == other.value
-
-    BUILD = "build"
-    TEST = "test"
-    DEPLOY = "deploy"
-    POST_DEPLOY = "postdeploy"
+class Stage:
+    name: str
+    icon: str
```
koeves, line 56: Maybe the icon should be optional

src/mpyl/project.py
```diff
@@ -134,12 +129,12 @@ class StageSpecificProperty(Generic[T]):
...
+    def for_stage(self, stage: str) -> Optional[T]:
+        if stage == "build":
             return self.build
-        if stage == Stage.TEST:
+        if stage == "test":
             return self.test
-        if stage == Stage.DEPLOY:
+        if stage == "deploy":
             return self.deploy
         return self.postdeploy
```
koeves, line 139: Hm these are still hardcoded.. But I don't think we can do much about it, maybe if we refactor this class to use a dictionary and then extract the properties by dict.get(stage)

SamTheisens: That would be possible, but the biggest issue I see is that we can't make the schema dynamic.

Jorg88, line 139: Shouldn't we throw instead of returning postdeploy as default in case a stage gets passed that isn't implemented (yet)?

src/mpyl/build.py
```diff
@@ -125,22 +119,20 @@ def run_mpyl(
...
+    stages: list[Stage],
+    build_all: bool,
 ) -> dict[Stage, set[Project]]:
     project_paths = repo.find_projects()
     all_projects = set(
         map(lambda p: load_project(Path(""), Path(p), False), project_paths)
     )
 
     if build_all:
-        return {
```
Jorg88, line 136: I think we can already decide on which approach we want to use between this and https://github.com/Vandebron/mpyl/pull/281/files#diff-5f8bc6ee933a7bacafed256143948c5c0b696a67bb5cddab95f2e0f3578f9dbaR138-R153 . There is no merge conflict except in the picked approach. Here the stages get passed and in my pr they are still hardcoded from a function, but for the rest the only difference is the way in which we coded it.

Maybe @koeves can vote on which way he likes most 😉

src/mpyl/reporting/formatting/markdown.py
```diff
@@ -64,15 +64,7 @@ def __to_oneliner(result: list[StepResult], plan: set[Project]) -> str:
 
 
 def stage_to_icon(stage: Stage):
```
Jorg88, line 66: this function is redundant now if you ask me

src/mpyl/stages/discovery.py
```diff
@@ -71,7 +72,7 @@ def _are_invalidated(
 
 
 def find_invalidated_projects_for_stage(
-    all_projects: set[Project], stage: Stage, change_history: list[Revision]
+    all_projects: set[Project], stage: str, change_history: list[Revision]
```
Jorg88, line 75: Yeah adding an enum as noted above would definitely be an improvement imo, because `Stage` being a class in some places/functions and a `str` in others can be quite confusing. Especially when used almost next to eachother.

tests/cli/commands/test_build.py
```diff
@@ -61,12 +63,17 @@ def test_run_build_with_plan_should_execute_successfully(self):
         run_properties = RUN_PROPERTIES
 
         projects = [get_minimal_project()]
-        run_plan = {Stage.BUILD: projects, Stage.TEST: projects, Stage.DEPLOY: projects}
+        run_plan = {
+            TestStage.build(): projects,
+            TestStage.test(): projects,
+            TestStage.deploy(): projects,
```
Jorg88, line 69: Not related to this pr, but shouldn't we add postdeploy to `TestStage` for this test and `tests/reporting/__init__.py` below?

src/mpyl/build.py
koeves: I thought we changed that to `[s for s in Stage]` or something like that in your PR. It should be possible to extract the stages in order from an Enum, that way it will be correct even if we add a new stage

Jorg88: True, I don't know what happened with that change. I'll see if I can implement it

src/mpyl/project.py
```diff
@@ -134,12 +129,12 @@ class StageSpecificProperty(Generic[T]):
     deploy: Optional[T]
     postdeploy: Optional[T]
 
-    def for_stage(self, stage: Stage) -> Optional[T]:
-        if stage == Stage.BUILD:
+    def for_stage(self, stage: str) -> Optional[T]:
+        if stage == "build":
```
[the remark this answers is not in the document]
SamTheisens: We can't have fully dynamic stages if we stick to enums. I'd like to give people the option to define their own stages eventually.

src/mpyl/build.py
SamTheisens: The order of stages is defined in `run_properties.yml`

src/mpyl/steps/models.py
```diff
@@ -102,8 +104,15 @@ def for_local_run(
             versioning=VersioningProperties(revision, branch, 123, tag),
             config=config,
             console=ConsoleProperties("INFO", True, 130),
+            stages=[],
         )
 
+    def stage_for(self, stage_name: str) -> Stage:
```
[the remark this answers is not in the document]
SamTheisens: `to_stage`? It's giving you the `Stage` enum for a string.

src/mpyl/stages/discovery.py
SamTheisens: https://github.com/Vandebron/mpyl/pull/255#discussion_r1359759603

tests/cli/commands/test_build.py
SamTheisens: What would be the goal? 
I think this emphasises that not every pipeline will use all of the stages.

src/mpyl/project.py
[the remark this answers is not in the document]
Jorg88: I'm just suggesting to replace the current hardcoded strings with an enum, that's not blocking anything or doing anything different than the current solution?

And if we do end up making it fully dynamic with stages coming from the config or something, then it's easier to find and replace the enum (values) compared to hardcoded strings.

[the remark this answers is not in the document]
SamTheisens: Yeah, but as a class though (https://github.com/Vandebron/mpyl/pull/255/files#diff-8ab9ddcdb8fee93bb0c9016ade42a4f20cb40bcfa5fd27e03d014c3c5725a542R54-R56) not an enum. 
So we should convert the stage string (from the cli for example) to a `Stage` object ASAP and then pass that around.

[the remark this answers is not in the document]
Jorg88: There are 2 different types though, there is the passed around dynamic stage that comes from the config/cli and (currently) is a string turned into a class. This is the one you are talking about. I'm talking about the strings that are used to say `if passed_around_stage.name == "build": do this specific thing`. In my head we can replace this comparison string "build" with an enum, since it's done multiple times for multiple stages. 

Anyway, it's not a big deal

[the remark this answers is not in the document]
SamTheisens: How can stages encoded in an enum be dynamic? We can't introduce a new stage purely through altering the config. If we use the class stage, we can create introduce as many different stages as we want.

Jorg88 (approved):

PR #262, opened 2023-09-27
[TECH-625] Inject target service name to match

oschrenk:
branch: fix/TECH-625-fix-service-monitor


oschrenk:
Ah. Didn't see 
https://github.com/Vandebron/mpyl/pull/261

koeves:
Ah nice, we should merge this PR instead so you become a contributor as well :)

PR #266, opened 2023-09-28
feature/TECH 633 docker build arguments

koeves:
Parse docker build arguments from project yaml

----
### 📕 [TECH-633](https://vandebron.atlassian.net/browse/TECH-633) Add docker build arguments danielkoves@vandebron.nl 

🏗️ Build [3](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-266/3/display/redirect) ✅ Successful, started by _Daniel Koves_  
🚀 *[cloudfront-service](https://cloudfront-service-266.test.nl/swagger/index.html)*, *example-dagster-user-code*, *job*, *kong-sync*, *[nodeservice](https://nodeservice-266.test.nl/swagger/index.html)*, *[sbtservice](https://sbtservice-266.test.nl/swagger/index.html)*, *sparkJob*  

[TECH-633]: https://vandebron.atlassian.net/browse/TECH-633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #268, opened 2023-09-29
Revert "feature/TECH 633 docker build arguments"

koeves:
Reverts Vandebron/mpyl#266

### DIFF
--- a/src/mpyl/steps/build/dockerbuild.py
+++ b/src/mpyl/steps/build/dockerbuild.py
@@ -20,7 +20,7 @@ to a folder named `$WORKDIR/target/test-reports/`.
 
 from logging import Logger
 
-from .post_docker_build import AfterBuildDocker
+from .docker_after_build import AfterBuildDocker
 from .. import Step, Meta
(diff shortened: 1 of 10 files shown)


src/mpyl/steps/build/dockerbuild.py
```diff
@@ -71,32 +71,22 @@ def execute(self, step_input: Input) -> Output:
...
-            build_args=build_args,
         )
         artifact = input_to_artifact(
             ArtifactType.DOCKER_IMAGE, step_input, spec=DockerImageSpec(image=image_tag)
         )
 
+        with open(".dockerignore", "w+", encoding="utf-8") as ignore_file:
+            contents = "\n".join(DOCKER_IGNORE_DEFAULT)
+            ignore_file.write(contents)
+
```
koeves, line 89: @SamTheisens why was the dockerignore part after the build?

SamTheisens: Yeah, good question. IIRC it was necessary to avoid the docker layers getting invalidated every build. But I think it's a good idea to validate that assumption.

PR #271, opened 2023-10-03
[TECH-639] Merge the available kubectl configs into one temporary con…

Jorg88:
…fig file to use in the docker container

branch: feature/TECH-639-account-for-multiple-kubeconfig-files


koeves (approved):
Looks good 👍

PR #278, opened 2023-10-06
feature/TECH 654 pass secrets to docker build

koeves:
Be able to specify credentials to be used in a Dockerfile, which are resolved from env vars

```yaml
build:
  args:
    plain:
      - key: SOME_ENV
        test: "Test"
        acceptance: "Acceptance"
        production: "Production"
    credentials:
      - key: SOME_CREDENTIAL
        id: SOME_CREDENTIAL
```

----
### 📕 [TECH-654](https://vandebron.atlassian.net/browse/TECH-654) Pass secrets to docker build danielkoves@vandebron.nl 
email connector proxy

msp

🏗️ Build [4](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-278/4/display/redirect) ✅ Successful, started by _Daniel Koves_  
🚀 *[cloudfront-service](https://cloudfront-service-278.test.nl/swagger/index.html)*, *example-dagster-user-code*, *job*, *kong-sync*, *[nodeservice](https://nodeservice-278.test.nl/swagger/index.html)*, *[sbtservice](https://sbtservice-278.test.nl/swagger/index.html)*, *sparkJob*  

[TECH-654]: https://vandebron.atlassian.net/browse/TECH-654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/mpyl/project.py
+++ b/src/mpyl/project.py
@@ -128,7 +128,7 @@ class KeyValueRef:
 
 
 @dataclass(frozen=True)
-class KeyId:
+class EnvCredential:
     key: str
     secret_id: str
 
@@ -138,7 +138,7 @@ class KeyId:
         secret_id = values.get("id")
         if not key or not secret_id:
             raise KeyError("Credential must have a key and id set.")
-        return KeyId(key, secret_id)
+        return EnvCredential(key, secret_id)
 
 
 @dataclass(frozen=True)
@@ -417,13 +417,15 @@ class Docker:
 @dataclass(frozen=True)
 class BuildArgs:
     plain: list[KeyValueProperty]
-    credentials: list[KeyId]
+    credentials: list[EnvCredential]
 
     @staticmethod
     def from_config(values: dict):
         return BuildArgs(
             plain=list(map(KeyValueProperty.from_config, values.get("plain", []))),
-            credentials=list(map(KeyId.from_config, values.get("credentials", []))),
+            credentials=list(
+                map(EnvCredential.from_config, values.get("credentials", []))
+            ),
         )
 
 
--- a/src/mpyl/project.py
+++ b/src/mpyl/project.py
@@ -130,15 +130,15 @@ class KeyValueRef:
 @dataclass(frozen=True)
 class KeyId:
     key: str
-    id: str
+    secret_id: str
 
     @staticmethod
     def from_config(values: dict):
         key = values.get("key")
-        id = values.get("id")
-        if not key or not id:
-            raise KeyError("Credental must have a key and id set.")
-        return KeyId(key, id)
+        secret_id = values.get("id")
+        if not key or not secret_id:
+            raise KeyError("Credential must have a key and id set.")
+        return KeyId(key, secret_id)
 
 
 @dataclass(frozen=True)
--- a/src/mpyl/project.py
+++ b/src/mpyl/project.py
@@ -128,6 +128,20 @@ class KeyValueRef:
 
 
 @dataclass(frozen=True)
+class KeyId:
+    key: str
+    id: str
+
+    @staticmethod
+    def from_config(values: dict):
+        key = values.get("key")
+        id = values.get("id")
+        if not key or not id:
+            raise KeyError("Credental must have a key and id set.")
+        return KeyId(key, id)
+
+
+@dataclass(frozen=True)
 class StageSpecificProperty(Generic[T]):
     build: Optional[T]
     test: Optional[T]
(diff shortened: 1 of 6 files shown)


src/mpyl/project.py
```diff
@@ -127,6 +127,20 @@ def from_config(values: dict):
         )
 
 
+@dataclass(frozen=True)
+class KeyId:
```
Jorg88, line 131: why `KeyId` and not `Credentials` or something?

src/mpyl/steps/build/docker_build.py
```diff
@@ -80,6 +80,14 @@ def execute(self, step_input: Input) -> Output:
                 arg.key: arg.get_value(step_input.run_properties.target)
                 for arg in step_input.project.build.args.plain
             }
+            | {
+                k: v
+                for (k, v) in {
+                    arg.key: os.getenv(arg.secret_id)
```
Jorg88, line 86: shouldn't we do some sort of check that it can find the secret in the env vars?

tests/projects/service/deployment/project.yml
```diff
@@ -13,6 +13,9 @@ build:
         test: "Test"
         acceptance: "Acceptance"
         production: "Production"
+    credentials:
+      - key: SOME_CREDENTIAL
+        id: SOME_CREDENTIAL
```
Jorg88, line 18: is there no test where we can set an env value and check if it actually shows up?

src/mpyl/project.py
koeves: Sure

src/mpyl/steps/build/docker_build.py
koeves: Yeah maybe that's better, I'll just throw if an env var is not set

tests/projects/service/deployment/project.yml
koeves: I was trying to think, but I don't think so.. With my changes to throw in case the env var is not set, at least the IT tests should fail

SamTheisens:
> Be able to specify credentials to be used in a Dockerfile, which are resolved from env vars
> 
> ```yaml
> build:
>   args:
>     plain:
>       - key: SOME_ENV
>         test: "Test"
>         acceptance: "Acceptance"
>         production: "Production"
>     credentials:
>       - key: SOME_CREDENTIAL
>         id: SOME_CREDENTIAL
> ```

What is the difference between plain and credential? Aren't both of them env vars?

koeves:
> > Be able to specify credentials to be used in a Dockerfile, which are resolved from env vars
> > ```yaml
> > build:
> >   args:
> >     plain:
> >       - key: SOME_ENV
> >         test: "Test"
> >         acceptance: "Acceptance"
> >         production: "Production"
> >     credentials:
> >       - key: SOME_CREDENTIAL
> >         id: SOME_CREDENTIAL
> > ```
> 
> What is the difference between plain and credential? Aren't both of them env vars?

plain a TargetProperty, so you could set them based on the environment and they are plain text. Credentials (in this version) are taken from the environment. In MPL they were used to get credentials by ID (by using `withCredentials`), now the workaround in MPyL is to define these `SOME_CREDENTIAL`s in the jenkinsfile's environment like SOME_CREDENTIAL = credential('etc') and then we'll have an exported SOME_CREDENTIAL to be passed on to docker build.

SamTheisens:
> > > Be able to specify credentials to be used in a Dockerfile, which are resolved from env vars
> > > ```yaml
> > > build:
> > >   args:
> > >     plain:
> > >       - key: SOME_ENV
> > >         test: "Test"
> > >         acceptance: "Acceptance"
> > >         production: "Production"
> > >     credentials:
> > >       - key: SOME_CREDENTIAL
> > >         id: SOME_CREDENTIAL
> > > ```
> > 
> > 
> > What is the difference between plain and credential? Aren't both of them env vars?
> 
> plain a TargetProperty, so you could set them based on the environment and they are plain text. Credentials (in this version) are taken from the environment. In MPL they were used to get credentials by ID (by using `withCredentials`), now the workaround in MPyL is to define these `SOME_CREDENTIAL`s in the jenkinsfile's environment like SOME_CREDENTIAL = credential('etc') and then we'll have an exported SOME_CREDENTIAL to be passed on to docker build.

What is the use case for this story? How are you planning to deal with invalidation? In principle it's possible now that you build a different image for acceptance than for production.

koeves:
> What is the use case for this story? How are you planning to deal with invalidation? In principle it's possible now that you build a different image for acceptance than for production.

We have the email connector proxy services which are just a dockerfile that copies a binary from s3 (for that we need the credentials now because platform made it so) and runs this binary.

How does invalidation come into play here?

Btw previously we _did_ build different images per environment. Currently the images are the same only the "command"s are different

PR #283, opened 2023-10-10
Move YAML parser back to default settings

### LINKED ISSUE #282: Helm chart templates are created with incorrect YAML strings
MPyL currently reads YAML files using `YAML(typ="unsafe")` (like [here](https://github.com/Vandebron/mpyl/blob/6f8e9a29be9c8991183d1240b9d00cc8096ed293/src/mpyl/project.py#L568) and [here](https://github.com/Vandebron/mpyl/blob/6f8e9a29be9c8991183d1240b9d00cc8096ed293/src/mpyl/project.py#L593)) , and then writes them back to disk using `YAML()` (which defaults to `typ="rt"`) everywhere else (most notably [here](https://github.com/Vandebron/mpyl/blob/6f8e9a29be9c8991183d1240b9d00cc8096ed293/src/mpyl/steps/deploy/k8s/resources/__init__.py#L16), where the chart templates are written back to disk). This creates some inconsistencies with block scalars, and this eventually is causing broken deployments due to invalid Helm chart templates.

A small demonstration:
```
yaml_str = """
    folded_block_scalar: >-
        1 first line
        2 second line
        3 a super duper really really really really really really really long line
        4 another super duper really really really really really really really long line
    literal_block_scalar: |-
        1 first line
        2 second line
        3 a super duper really really really really really really really long line
        4 another super duper really really really really really really really long line
"""
```

Reading and writing with the default settings:
```
yaml = YAML()
values = yaml.load(yaml_str)
print(yaml_to_string(values, yaml))
```

... will print, as expected, the exact same representation:
```
folded_block_scalar: >-
  1 first line
  2 second line
  3 a super duper really really really really really really really long line
  4 another super duper really really really really really really really long line
literal_block_scalar: |-
  1 first line
  2 second line
  3 a super duper really really really really really really really long line
  4 another super duper really really really really really really really long line
```

---

Using the YAML parser with the unsafe setting:
```
yaml = YAML(typ="unsafe")
values = yaml.load(yaml_str)
print(yaml_to_string(values, yaml))      
```

.. will start showing some interesting results:
```
{folded_block_scalar: 1 first line 2 second line 3 a super duper really really really
    really really really really long line 4 another super duper really really really
    really really really really long line, literal_block_scalar: '1 first line

    2 second line

    3 a super duper really really really really really really really long line

    4 another super duper really really really really really really really long line'}

```

This looks very different than the original representation, however both strings actually mean the exact same thing:
* `folded_block_scalar` was _folded_ to a plain string as expected
* `literal_block_scalar` was transformed into a single quoted string, which has its own set of rules but mean

---
Really interesting things happen when we mix and match, though:
```
values = YAML(typ="unsafe").load(yaml_str)
print(yaml_to_string(values, YAML()))
```
... produces the output:
```
folded_block_scalar: 1 first line 2 second line 3 a super duper really really really
  really really really really long line 4 another super duper really really really
  really really really really long line
literal_block_scalar: "1 first line\n2 second line\n3 a super duper really really\
  \ really really really really really long line\n4 another super duper really really\
  \ really really really really really long line"
```
`folded_block_scalar` actually looks fine, but the `literal_block_scalar` looks quite broken. 

It's now represented as a double quoted string, and for some reason backwards slashes (`\`) are used to split the string into multiple lines. This is causing issues when applied to values with existing escape characters, causing them to be put together and parsed as `\\` - breaking the application of Helm chart templates.

ptab:
Fixes #282

I looked back in history for clues as to why this `typ="unsafe"` is needed, and apparently it was done [for performance reasons](https://github.com/Vandebron/mpyl/commit/2b7bee0b9f81841d7b47691b74b12312f9556102#diff-8ab9ddcdb8fee93bb0c9016ade42a4f20cb40bcfa5fd27e03d014c3c5725a542R348).
My suggestion is to disable this setting since it seems to be causing more harm than benefit. 

----
 # ⚠️ Could not find ticket corresponding to `fix/move-yaml-settings-back-to-default. Does your branch name follow the correct pattern?

### DIFF
--- a/src/mpyl/project.py
+++ b/src/mpyl/project.py
@@ -20,7 +20,7 @@ import traceback
 from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
-from typing import Optional, TypeVar, Any, TextIO, List
+from typing import Optional, TypeVar, Any, List
 
 import jsonschema
 from mypy.checker import Generic
@@ -559,13 +559,12 @@ class Project:
         )
 
 
-def validate_project(file: TextIO) -> dict:
+def validate_project(yaml_values: dict) -> dict:
     """
     :file the file to validate
     :return: the validated schema
     :raises `jsonschema.exceptions.ValidationError` when validation fails
     """
-    yaml_values = YAML().load(file)
     template = pkgutil.get_data(__name__, "schema/project.schema.yml")
     if not template:
         raise ValueError("Schema project.schema.yml not found in package")
@@ -575,7 +574,11 @@ def validate_project(file: TextIO) -> dict:
 
 
 def load_project(
-    root_dir: Path, project_path: Path, strict: bool = True, log: bool = True
+    root_dir: Path,
+    project_path: Path,
+    strict: bool = True,
+    log: bool = True,
+    safe: bool = False,
 ) -> Project:
     """
     Load a `project.yml` to `Project` data class
@@ -583,13 +586,17 @@ def load_project(
     :param project_path: relative path from `root_dir` to the `project.yml`
     :param strict: indicates whether the schema should be validated
     :param log: indicates whether problems should be logged as warning
+    :param safe: indicates that correctness should be prioritized over speed. Safe loading is important
+    when the values possibly end up in artifacts
     :return: `Project` data class
     """
     log_level = logging.WARNING if log else logging.DEBUG
     with open(root_dir / project_path, encoding="utf-8") as file:
         try:
             start = time.time()
-            yaml_values = validate_project(file) if strict else YAML().load(file)
+            yaml_values = YAML(typ=None if safe else "unsafe").load(file)
+            if strict:
+                validate_project(yaml_values)
             project = Project.from_config(yaml_values, project_path)
             logging.debug(
                 f"Loaded project {project.path} in {(time.time() - start) * 1000} ms"
--- a/src/mpyl/project.py
+++ b/src/mpyl/project.py
@@ -565,7 +565,7 @@ def validate_project(file: TextIO) -> dict:
     :return: the validated schema
     :raises `jsonschema.exceptions.ValidationError` when validation fails
     """
-    yaml_values = YAML(typ="unsafe").load(file)
+    yaml_values = YAML().load(file)
     template = pkgutil.get_data(__name__, "schema/project.schema.yml")
     if not template:
         raise ValueError("Schema project.schema.yml not found in package")
@@ -589,9 +589,7 @@ def load_project(
     with open(root_dir / project_path, encoding="utf-8") as file:
         try:
             start = time.time()
-            yaml_values = (
-                validate_project(file) if strict else YAML(typ="unsafe").load(file)
-            )
+            yaml_values = validate_project(file) if strict else YAML().load(file)
             project = Project.from_config(yaml_values, project_path)
             logging.debug(
                 f"Loaded project {project.path} in {(time.time() - start) * 1000} ms"
--- a/src/mpyl/build.py
+++ b/src/mpyl/build.py
@@ -29,6 +29,7 @@ def get_build_plan(
     repo: Repository,
     run_properties: RunProperties,
     cli_parameters: MpylCliParameters,
+    safe_load_projects: bool,
 ) -> RunResult:
     tag = run_properties.versioning.tag
     if tag:
@@ -46,7 +47,7 @@ def get_build_plan(
     logger.debug(f"Changes: {changes}")
 
     projects_per_stage: dict[Stage, set[Project]] = find_build_set(
-        repo, changes, cli_parameters.all
+        repo, changes, cli_parameters.all, safe_load_projects
     )
     return RunResult(
         run_properties=run_properties,
@@ -86,6 +87,7 @@ def run_mpyl(
                 repo=repo,
                 run_properties=run_properties,
                 cli_parameters=cli_parameters,
+                safe_load_projects=True,
             )
 
             if not run_plan.run_plan.items():
@@ -125,11 +127,23 @@ def run_mpyl(
 
 
 def find_build_set(
-    repo: Repository, changes_in_branch: list[Revision], build_all: bool
+    repo: Repository,
+    changes_in_branch: list[Revision],
+    build_all: bool,
+    safe_load_projects: bool,
 ) -> dict[Stage, set[Project]]:
     project_paths = repo.find_projects()
     all_projects = set(
-        map(lambda p: load_project(Path(""), Path(p), False), project_paths)
+        map(
+            lambda p: load_project(
+                root_dir=Path(""),
+                project_path=Path(p),
+                strict=False,
+                log=True,
+                safe=safe_load_projects,
+            ),
+            project_paths,
+        )
     )
 
     if build_all:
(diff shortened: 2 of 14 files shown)


SamTheisens:
> Fixes #282
> 
> I looked back in history for clues as to why this `typ="unsafe"` is needed, and apparently it was done [for performance reasons](https://github.com/Vandebron/mpyl/commit/2b7bee0b9f81841d7b47691b74b12312f9556102#diff-8ab9ddcdb8fee93bb0c9016ade42a4f20cb40bcfa5fd27e03d014c3c5725a542R348). My suggestion is to disable this setting since it seems to be causing more harm than benefit.

The performance improvement form using 'unsafe' is _significant_. Since all projects are loaded for any build operation, the cli becomes sluggish if we load the `project.yml` safely.
The `project.yml` that are thus loaded are never serialized back to to yaml, so I don't understand how the scenario describe could happen.
Do you have a real life example?

ptab:
> The project.yml that are thus loaded are never serialized back to to yaml, so I don't understand how the scenario describe could happen.

I'm new to MPyL so maybe I'm not explaining the situation in a clear way.
As far as I can tell, the `project.yml` is [loaded from a YAML file into an in-memory representation](https://github.com/Vandebron/mpyl/blob/6f8e9a29be9c8991183d1240b9d00cc8096ed293/src/mpyl/project.py#L592-L595):
```python
yaml_values = (
    validate_project(file) if strict else YAML(typ="unsafe").load(file)
)
project = Project.from_config(yaml_values, project_path)
```

During the deployment phase, that same in-memory representation is [first converted to a different representation](https://github.com/Vandebron/mpyl/blob/6f8e9a29be9c8991183d1240b9d00cc8096ed293/src/mpyl/steps/deploy/kubernetes.py#L59):
```python
chart = to_service_chart(builder)
```

And `chart` is then eventually [serialized back to YAML and written to disk](https://github.com/Vandebron/mpyl/blob/main/src/mpyl/steps/deploy/k8s/__init__.py#L241-L243):
```python
chart_path = write_helm_chart(
    logger, chart, Path(project.target_path), run_properties, release_name
)
```
This serialization is done [using a default (not unsafe) instance of YAML](https://github.com/Vandebron/mpyl/blob/6f8e9a29be9c8991183d1240b9d00cc8096ed293/src/mpyl/steps/deploy/k8s/resources/__init__.py#L16), triggering #282:
```python
yaml = YAML()
``` 

The location of the Helm charts on disk (`chart_path`) is then [used to trigger installation of the Helm charts](https://github.com/Vandebron/mpyl/blob/main/src/mpyl/steps/deploy/k8s/__init__.py#L275):
```python
return helm.install(
    logger,
    chart_path,
    dry_run,
    release_name,
    namespace,
    rancher_config.context,
    delete_existing,
)
```
And because the generated YAML files will contain a different string representation than what was present in the original `project.yml` (as shown in #282), this will sometimes cause the Helm deployment to fail.

--- 

I'll try to create a test case to visualize this more explicitly, but I hope this message explains the issue we're seeing.

SamTheisens:
> > The project.yml that are thus loaded are never serialized back to to yaml, so I don't understand how the scenario describe could happen.
> 
> I'm new to MPyL so maybe I'm not explaining the situation in a clear way. As far as I can tell, the `project.yml` is [loaded from a YAML file into an in-memory representation](https://github.com/Vandebron/mpyl/blob/6f8e9a29be9c8991183d1240b9d00cc8096ed293/src/mpyl/project.py#L592-L595):
> 
> ```python
> yaml_values = (
>     validate_project(file) if strict else YAML(typ="unsafe").load(file)
> )
> project = Project.from_config(yaml_values, project_path)
> ```
> 
> During the deployment phase, that same in-memory representation is [first converted to a different representation](https://github.com/Vandebron/mpyl/blob/6f8e9a29be9c8991183d1240b9d00cc8096ed293/src/mpyl/steps/deploy/kubernetes.py#L59):
> 
> ```python
> chart = to_service_chart(builder)
> ```
> 
> And `chart` is then eventually [serialized back to YAML and written to disk](https://github.com/Vandebron/mpyl/blob/main/src/mpyl/steps/deploy/k8s/__init__.py#L241-L243):
> 
> ```python
> chart_path = write_helm_chart(
>     logger, chart, Path(project.target_path), run_properties, release_name
> )
> ```
> 
> This serialization is done [using a default (not unsafe) instance of YAML](https://github.com/Vandebron/mpyl/blob/6f8e9a29be9c8991183d1240b9d00cc8096ed293/src/mpyl/steps/deploy/k8s/resources/__init__.py#L16), triggering #282:
> 
> ```python
> yaml = YAML()
> ```
> 
> The location of the Helm charts on disk (`chart_path`) is then [used to trigger installation of the Helm charts](https://github.com/Vandebron/mpyl/blob/main/src/mpyl/steps/deploy/k8s/__init__.py#L275):
> 
> ```python
> return helm.install(
>     logger,
>     chart_path,
>     dry_run,
>     release_name,
>     namespace,
>     rancher_config.context,
>     delete_existing,
> )
> ```
> 
> And because the generated YAML files will contain a different string representation than what was present in the original `project.yml` (as shown in #282), this will sometimes cause the Helm deployment to fail.
> 
> I'll try to create a test case to visualize this more explicitly, but I hope this message explains the issue we're seeing.

Do you have een example `project.yml` where this happens?

ptab:
Here's the one that is currently triggering this issue:
```yaml
name: 'application'
stages:
  build: 'Sbt Build'
  test: 'Sbt Test'
  deploy: 'Kubernetes Deploy'
deployment:
  kubernetes:
    metrics:
      alerts:
        - name: "the name of the alert"
          expr: "a prometheus expression"
          forDuration: 0m # Fire instantly
          severity: signup_alerting_once
          description: |-
            **{{ $value }}** new signup failures with reason `{{ $labels.reason }}`.
            You can find more details in: <unrelated link>
```

Note: I removed all other unrelated fields, hopefully nothing is mandatory.

ptab:
The previous `project.yml` will generate a Prometheus rule template like:
```yaml
spec:
  groups:
  - name: customers-prometheus-rule-group
    rules:
    - alert: "the name of the alert"
      annotations:
        description: "**{{ \"{{\" }} $value }}** new signup failures with reason `{{\
          \ \"{{\" }} $labels.reason }}`.\nYou can find more details in: <redacted link>"
```
Double-quoted strings are meant to render without newlines, and that will produce `\\` characters which Helm rejects.

src/mpyl/project.py
```diff
@@ -565,7 +565,7 @@ def validate_project(file: TextIO) -> dict:
     :return: the validated schema
     :raises `jsonschema.exceptions.ValidationError` when validation fails
     """
-    yaml_values = YAML(typ="unsafe").load(file)
+    yaml_values = YAML().load(file)
```
SamTheisens, line 568: This will slow matters down way too much. I would suggest to reload the project (safely) here only https://github.com/Vandebron/mpyl/blob/caf442e2df0d8be92ba33a652fd1e785a6839e7d/src/mpyl/build.py#L155

Because it's only in the context of building, deployment in particular that the yaml parsing is relevant.

SamTheisens: @ptab I made the safe load conditional here: https://github.com/Vandebron/mpyl/pull/283/commits/b9b3b94baa7f714c371be8d1d65fa3f9df613906

That should fix the issue, but routine operations like `mpyl build status` snappy.

ptab: Sounds like a good compromise, thanks for finding an alternative solution 👍

Jorg88 (approved):

PR #286, opened 2023-10-24
Fix the logging of the git commit check, the reporter and the runner

Jorg88:
----
 # ⚠️ Could not find ticket corresponding to `feature/TECH-XXX-debug-github-status-check. Does your branch name follow the correct pattern?


koeves (approved):
What exactly was the issue with the check?

Jorg88:
> What exactly was the issue with the check?

The old function had an if/else based on the run_id of the github pr. If you would call it the first time it would get and set the run_id, set the status check to "in progress" (as can be seen in the caching pr right now) and leave it there. Then you would need to call the function again for it to go into the other flow and set the status check to the result of the run.

This used to make sense in the old runner in which we would call it twice, first before the mpyl run and then after. The downside of the current flow is that the status check won't be "in_progress" while mpyl is running.

koeves (approved):
Still looks good!

PR #287, opened 2023-10-25
[TECH-494] Implement multiple deployments for one project

nik-kashi:
It recognizes the `project-override*.yml` files and merges them to the parent yml file. The rest is the same behavior as it was.

----
### 📕 [TECH-494](https://vandebron.atlassian.net/browse/TECH-494) Migrate project overrides nikkashi@vandebron.nl 
**What**

We currently do not support project yaml overrides in MPyL. It is also interfering with the namespace linkup linting step (because the project name is dynamic).

Teams indicated that they are going to migrate some project away from using these overrides, see: [2023-07-26+CI+CD+Guild](https://vandebron.atlassian.net/wiki/spaces/DIG/pages/95581484646441/2023-07-26+CI+CD+Guild) 

For the rest, come up with a plan on how to migrate them, talk to teams that have projects like that.

For argo, take a look at applicationset: [](https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/) 

----

**How**

There’s a main project yaml and override yamls. The override yaml should be attached to the main one which create a new project.

Current shortcomings:

* env vars need to be duplicated
* docker images are created multiple times (per override)

The overrides concern the `deployment` section of the project yaml.

Then, when constructing the build plan, the overrides would only be applied to the Deployment stage.

So we’ll have one run of build and test stages, then multiple projects will be deployed from it with different input and environment.

example build plan:

build: `ocpp`

test: `ocpp`

deployment: `ocpp-1}}, {{ocpp-2` , … => should use the same image that was created before

----

figure out a name for this (project merge)

create a schema → subset of the project yaml schema and validate the overridden projects

in MPL overrides were detected by a naming convention: `project-override-XXX.yml}}, the main project has a {{project.yaml`

----

Check out [kustomize](https://kustomize.io/), if it can be applied as after the deployment stage

🏗️ Build [2](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-287/2/display/redirect) ✅ Successful, started by _Nik Kashi_  
🚀 *[cloudfront-service](https://cloudfront-service-287.test.nl/swagger/index.html)*, *example-dagster-user-code*, *job*, *kong-sync*, *[nodeservice](https://nodeservice-287.test.nl/swagger/index.html)*, *ocpp-first*, *ocpp-second*, *[sbtservice](https://sbtservice-287.test.nl/swagger/index.html)*, *sparkJob*  

[TECH-494]: https://vandebron.atlassian.net/browse/TECH-494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/mpyl/project.py
```diff
@@ -534,9 +534,19 @@ def job(self) -> Job:
...
     @property
     def root_path(self) -> str:
-        return self.path.replace(Project.project_yaml_path(), "")
+        if self.path.endswith(".yml"):
+            try:
+                return str(Path(self.path).parents[1])
+            except IndexError:
+                return self.path
+        else:
+            return self.path
```
Jorg88, line 549:
```suggestion
        if self.path.endswith(".yml"):
            try:
                return str(Path(self.path).parents[1])
            except IndexError:
                pass
        return self.path
```

A bit more clean maybe?

src/mpyl/cli/projects.py
```diff
@@ -98,8 +98,7 @@ def shell_complete(self, ctx: click.Context, param, incomplete: str):
         )
         found_projects = repo.find_projects(incomplete)
         return [
-            CompletionItem(value=proj.replace(f"/{Project.project_yaml_path()}", ""))
-            for proj in found_projects
+            CompletionItem(value=str(Path(proj).parents[1])) for proj in found_projects
```
Jorg88, line 101: I see `.parents[1]` being used multiple times, maybe we should make another static function in `Project` to return the project root path from the full yaml path? That would make it a bit easier to maintain if the folder structure would ever change

src/mpyl/project.py
```diff
@@ -631,6 +658,24 @@ def load_project(
             raise
 
 
+def merge_dicts(child_yaml_values, parent_yaml_values):
```
Jorg88, line 661: same as above and don't forget to add some types here

src/mpyl/project.py
```diff
@@ -609,7 +634,9 @@ def load_project(
     with open(root_dir / project_path, encoding="utf-8") as file:
         try:
             start = time.time()
-            yaml_values = YAML(typ=None if safe else "unsafe").load(file)
+            child_yaml_values = YAML(typ=None if safe else "unsafe").load(file)
```
Jorg88, line 637: nit: keep the old name, you don't know if this is a 'child' yaml 😬

src/mpyl/project.py
```diff
@@ -588,6 +598,21 @@ def validate_project(yaml_values: dict) -> dict:
...
+    safe: bool = False,
+) -> Optional[dict]:
+    full_path = root_dir / project_path
+    if str(full_path).endswith(Project.project_yaml_path()):
+        return None
+    parent_project_path = full_path.parents[1] / Project.project_yaml_path()
+    if not parent_project_path.exists():
+        return None
+    with open(parent_project_path, encoding="utf-8") as file:
+        return YAML(typ=None if safe else "unsafe").load(file)
```
Jorg88, line 613:
```suggestion
def load_possible_parent(
    full_path: Path,
    safe: bool = False,
) -> Optional[dict]:
    parent_project_path = full_path.parents[1] / Project.project_yaml_path()
    if (
        os.path.basename(full_path) == Project.project_yaml_path()
        or not parent_project_path.exists()
    ):
        return None
    with open(parent_project_path, encoding="utf-8") as file:
        return YAML(typ=None if safe else "unsafe").load(file)
```

src/mpyl/cli/projects.py
nik-kashi: Addressed in https://github.com/Vandebron/mpyl/pull/287/commits/91bb7489d9cfdacc6f7855d6af138c8025478c99

src/mpyl/project.py
nik-kashi: Addressed in https://github.com/Vandebron/mpyl/pull/287/commits/91bb7489d9cfdacc6f7855d6af138c8025478c99

src/mpyl/project.py
nik-kashi: [Addresse](https://github.com/Vandebron/mpyl/pull/287/commits/91bb7489d9cfdacc6f7855d6af138c8025478c99)

src/mpyl/project.py
nik-kashi: https://github.com/Vandebron/mpyl/pull/287/commits/91bb7489d9cfdacc6f7855d6af138c8025478c99

src/mpyl/project.py
Jorg88: Hm, the naming is still wrong though? In 95% of the cases it's not a `child_yaml` but just a yaml right?

src/mpyl/project.py
Jorg88: that's missing the part of my suggestion where the `return None` is combined though

SamTheisens:
The possibility to customize `project.yml` via overlays was introduced in MPL to allow the _same artifact_ to be deployed with slightly different deployment parameters.

In hindsight, I think I would do it somewhat differently. To summarize, I think the current override mechanism is too powerful. It has too many degrees of freedom. 
Some pros and cons:

**Advantages**
 1. It's quite a simple mechanism. Compared to full fledged templating tool like HELM, it's much easier to reason about and predict the outcome.
 2. The mechanism is sufficiently powerful. 
 3. It's built in to MPyL, doesn't require 3rd party dependencies and can be fully (unit) tested 
 4. It's familiar to the people who have been using MPL

**Drawbacks**
 1. Since overlays allow you to transform any part of the `project.yml`, the sections that don't relate to the deployment stage can be altered too. This makes it theoretically possible to create a different artifact for acceptance than for production, via docker build arguments for example.
This is potentially quite dangerous. You may get different behavior in production than what you've tested in acceptance. It also undermines the detection of artifact invalidation. This means that we can't reliably reuse the artifacts built earlier, leading to (dramatically) increased build times.
 2. The mechanism may not be immediately obvious to new users. This can be mitigated, to an extent, by improving documentation. But if we'd use a standard tool like kustomize, we could benefit from their documentation and community support.  
 3. There is currently no schema for the overlay files, and I think they would very hard if not impossible to implement. This means that you don't have autocompletion while composing overlays. But importantly, the `project.yml` resulting from applying the overlays can still be validated against the schema. So this drawback is probably not more than a minor nuisance.
 

Taking these aspects into account, I would suggest to implement the override (maybe rename it to overlay) as is, with one restriction: it would apply to the _deployment_ section of the `project.yml` only. This would overcome drawback number 1, while still being able to support the current use cases. Please correct me if I'm wrong on the latter!

More generally, going forward, I would suggest to eliminate the use of HELM completely in favor of standard k8s manifests deployed with `kubectl`.
At this point the [problematic templating](https://anthonyhawkins.medium.com/is-python-the-perfect-json-yaml-templating-engine-c5c1b32418f6) part of HELM isn't used anyway. HELM is only used to deploy pre-rendered manifests, for historical reasons.
Since MPyL can already output k8s manifests, `kubectl` could be used instead of HELM. I see couple of advantages to this approach:

- We can reduce our footprint by eliminating a native dependency on a 3rd party tool. This also reduces mental overhead, because there's an entire tool and ecosystem that users no longer need to worry about.
- Kubectl is standard k8s, so not a big ask for people that want to deploy to a k8s cluster
- The rendered k8s manifests are very WYSIWIG. In contrast to HELM templates they relate 1-to-1 to the CRD yaml that you will encounter in k8s.
- It opens the door to using application sets in ArgoCD with kustomize overlays. That way the responsibilities of MPyL can be reduced. The ultimate responsibility of MPyL would be to build artifacts and render manifests. The deployment and customization of those manifests would be ArgoCD's, with the help of kustomize.

nik-kashi:
Thanks Sam for the feedback.
About inheriting the fields, I am not sure if your concern can be happened in practice. Because, every overridden project has a new name and its own deployments, and it will be deployed for all three environment. We can not override a stage of a project, but defining a new project(deployment) using the parent template

src/mpyl/project.py
nik-kashi: the `basename` can not be used here
The two if blocks are merged here:
https://github.com/Vandebron/mpyl/pull/287/commits/7e7a3c7ec43c20dfc33b926466c73d79d02f17ea

src/mpyl/project.py
nik-kashi: https://github.com/Vandebron/mpyl/pull/287/commits/7e7a3c7ec43c20dfc33b926466c73d79d02f17ea

SamTheisens:
> Thanks Sam for the feedback. About inheriting the fields, I am not sure if your concern can be happened in practice. Because, every overridden project has a new name and its own deployments, and it will be deployed for all three environment. We can not override a stage of a project, but defining a new project(deployment) using the parent template

Indeed, the `name` field _would_ need to be overlayed. But besides the name field, do you think it would be possible to _just_ override whatever is under the `deployment` field?

nik-kashi:
> > Thanks Sam for the feedback. About inheriting the fields, I am not sure if your concern can be happened in practice. Because, every overridden project has a new name and its own deployments, and it will be deployed for all three environment. We can not override a stage of a project, but defining a new project(deployment) using the parent template
> 
> Indeed, the `name` field _would_ need to be overlayed. But besides the name field, do you think it would be possible to _just_ override whatever is under the `deployment` field?

Addressed here https://github.com/Vandebron/mpyl/pull/287/commits/04f0423d3ccea0e7f579b3bd0021f3ad458a02e8

Jorg88 (approved):

koeves (approved):

PR #294, opened 2023-11-02
Fix project overlaying

nik-kashi:
----
 # ⚠️ Could not find ticket corresponding to `fix/port-mapping. Does your branch name follow the correct pattern?

### DIFF
--- a/src/mpyl/build.py
+++ b/src/mpyl/build.py
@@ -175,7 +175,13 @@ def run_mpyl(
                 console.log(f"Exception during build execution: {exc}")
                 console.print_exception()
 
-            console.print(Markdown(run_result_to_markdown(run_result)))
+            try:
+                console.print(Markdown(run_result_to_markdown(run_result)))
+            except Exception as exc:
+                console.log(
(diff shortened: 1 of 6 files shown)


src/mpyl/build.py
```diff
@@ -175,7 +175,14 @@ def run_mpyl(
                 console.log(f"Exception during build execution: {exc}")
                 console.print_exception()
 
-            console.print(Markdown(run_result_to_markdown(run_result)))
+            try:
```
Jorg88, line 178: I don't think this is the best solution, it just hides the actual problem? We also never had these markdown issues, so it must be something new or related to this pr which we should fix

nik-kashi: Agree, I'll work on it

nik-kashi: Addressed in : https://github.com/Vandebron/mpyl/pull/294/commits/abed1d80695b25ee47189b790a4ce2fd2fb4f96d

Jorg88 (approved):

koeves (approved):

PR #295, opened 2023-11-02
Feature/tech 610 implement argocd

Jorg88:
Task list:

- [x] Re-use caching push logic for our argo purposes
- [x] Rename config repo's as per https://github.com/Vandebron/mpyl/pull/295#discussion_r1381575376
- [x] Create pr out of branches using PyGithub
- [x] Check with Sam on pr comments and changed usage of manifest pushes

----

### 📕 [GUILD-610](https://vandebron.atlassian.net/browse/GUILD-610) Refactor MPyL to not deploy to K8S directly jorgpost@vandebron.nl 
Mpyl already supports creating the required manifest by setting

```yaml
kubernetes:
  deployAction: KubectlManifest
  outputPath: (optional){code}

in the `mpyl*config}}. The default {{outputPath}} is {{{projectname}/target.kubernetes/{project}.yml}}. The logic to create this yml is in the {{deploy_helm_chart}} function. This is also where the logic to push to the argo repo should be added, at the end of the {{if action == DeployAction.KUBERNETES*MANIFEST.value:` codeblock.

We can re-use the `Repository}} class and take some of the logic of the {{ArtifactsRepository}} class as an example to checkout {{main` in a temp folder, copy (overwrite) the build manifests to the correct folders (based on current deploy target) and commit & push the changes.

The folder structure in main should look like:

{code:none}root/
  project1/
    pr-xxxx/
    pr-xxxy/
    test/
    acce/
    prod/
  project2/
    etc/

```

This way the logic should not interfere with the local & remote caching or break the current `.mpyl` folder system.

🏗️ Build [8](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-295/8/display/redirect) ✅ Successful, started by _Jorg Post_  
🚀 *[cloudfront-service](https://cloudfront-service-295.test.nl/swagger/index.html)*, *example-dagster-user-code*, *job*, *kong-sync*, *[nodeservice](https://nodeservice-295.test.nl/swagger/index.html)*, *ocpp-first*, *ocpp-second*, *[sbtservice](https://sbtservice-295.test.nl/swagger/index.html)*, *sparkJob-first*, *sparkJob-second*  

[GUILD-610]: https://vandebron.atlassian.net/browse/GUILD-610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/mpyl/steps/deploy/k8s/kubectl.py
```diff
@@ -0,0 +1,85 @@
...
+                manifest_path.name,
+            )
+            shutil.copytree(
+                src=manifest_path.parent,
+                dst=new_file_path.parent,
+                dirs_exist_ok=True,
+            )
+
+            #  validate manifest
+            validate_command = (
```
koeves, line 60: Maybe good to think a bit on this approach. Once we're fully on argo, I think it would make sense to get rid of as much of the k8s related concepts from mpyl as possible. No helm, no rancher config etc.

src/mpyl/steps/deploy/k8s/kubectl.py
```diff
@@ -0,0 +1,85 @@
...
+    manifest_path: Path,
+) -> Output:
+    git_config = step_input.run_properties.config["vcs"]["argoRepository"]
+    if not git_config:
+        raise ValueError("No argocd repository configured")
+    argocd_repo_config = RepoConfig.from_git_config(git_config=git_config)
+    with TemporaryDirectory() as tmp_repo_dir:
+        with Repository.from_clone(
+            config=argocd_repo_config, repo_path=Path(tmp_repo_dir)
+        ) as artifact_repo:
```
koeves, line 28: By the way, we use the name artifact_repo for both the caching and the argo ones. I would name the variables more consistently

src/mpyl/steps/deploy/k8s/kubectl.py
Jorg88: I don't see any harm in using kubectl? It can be used for so much more than just deploying. Besides, how else do we make sure we push valid manifests to the *main* branch of the argo repo?

src/mpyl/steps/deploy/k8s/kubectl.py
```diff
@@ -0,0 +1,82 @@
...
+) -> Output:
+    git_config = step_input.run_properties.config["vcs"]["argoRepository"]
+    if not git_config:
+        raise ValueError("No argocd repository configured")
+    argocd_repo_config = RepoConfig.from_git_config(git_config=git_config)
+    with TemporaryDirectory() as tmp_repo_dir:
+        with Repository.from_clone(
+            config=argocd_repo_config, repo_path=Path(tmp_repo_dir)
+        ) as argo_repo:
+            branch = "feature/TECH-610-implement-argocd-2"  # This can be main later
```
Jorg88, line 30: Reminder to change this before merging

src/mpyl/steps/deploy/k8s/kubectl.py
ptab: I agree with @koeves 👍 

Instead of pushing directly to the `main` branch, we can create a PR, configure an action to automatically validate, lint, and whatever else we need, and automatically merge to main when green. I used this exact same approach in the past and it worked _really well_.

This also solves the problem of race conditions with multiple deployments: using the current approach of pushing to `main` will work, but we'll immediately start seeing a lot of 
```
error: failed to push some refs to '<repository>'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

once every project starts deploying at same time.

src/mpyl/schema/mpyl_config.schema.yml
```diff
@@ -176,6 +176,9 @@ definitions:
       artifactRepository:
         description: "The repository where the build artifacts are persisted"
         "$ref": "#/definitions/Git"
+      argoRepository:
+        description: "The argocd repository that contains the kubernetes manifests"
+        "$ref": "#/definitions/Git"
```
SamTheisens, line 181: Artifact repository was meant to cover both build cache and manifests. I understand you need to be able to store them in different repositories.
For the naming perhaps you could follow the choices here https://github.com/Vandebron/mpyl/blob/32476643940c8daa4bc38658170e2a937c24ff8e/src/mpyl/cli/build.py#L489

SamTheisens (changes requested):

src/mpyl/steps/deploy/k8s/kubectl.py
SamTheisens: I forced a schema error https://github.com/SamTheisens/mpyl-example-argocd/pull/3/commits/53d5e9bfecfb0277531ad423465652c2c2a20d81

Not the clearest error, but the deployment fails:
Screenshot 2023-11-03 at 20 23 15

src/mpyl/steps/deploy/k8s/kubectl.py
```diff
@@ -0,0 +1,67 @@
...
+            shutil.copytree(
+                src=manifest_path.parent,
+                dst=new_file_path.parent,
+                dirs_exist_ok=True,
+            )
+
+            if argo_repo.has_changes:
+                argo_repo.stage(".")
+                argo_repo.commit(
+                    f"Committing new manifest for project {step_input.project.name} on environment {folder_name}"
```
ptab, line 53: Is it possible to add either the branch name or the PR number to the commit message?
Something like:
* _Deploy project {step_input.project.name} to environment {folder_name} from master branch_
* _Deploy project {step_input.project.name} to environment {folder_name} from PR-123_ 

This would make the repository history easier to parse.

ptab: Thinking a bit more about this, I think it would be wise to make deployments atomic, i.e. changes to all applications involved in the same PR or tag to be applied to this repository at the same time (either in the same commit, or in multiple commits in a branch) instead of in multiple operations.
Like [Sam mentioned in a different thread](https://github.com/Vandebron/mpyl/pull/295#discussion_r1381565092), this will make rollbacks trivial as it will mean simply reverting a single commit.

SamTheisens: I'll try to clear up 

***Production*** The manifest repo is a mirror image of your infrastructure. The manifests in this [folder](https://github.com/SamTheisens/mpyl-example-argocd/tree/main/manifests/projects) are mapped nearly 1-to-1 to applications via [this generator](https://github.com/SamTheisens/mpyl-example-argocd/blob/main/argo/monorepo_application.yaml).
They end up in ArgoCD (and the K8S cluster) as follows:
Screenshot 2023-11-04 at 09 28 00
The generator point to `HEAD` of `main`. Any revision pushed to main results in a deployment of one or more applications, depending which applications' manifests were altered.
So ArgoCD deployments are per **repo**, not per application.

***Acceptance*** should be as much as possible identical to production. The acceptance and production generators cannot point to the same folder with manifests, because the manifests will differ (slightly). You could use separate a folder for `test`, `acceptance` and `production`. 

***PR*** for each pull request in your _source code_ repo a counterpart is created in the _artifact_ repo. Only the applications that were altered will receive alterations to the manifests. These manifests are deployed into a namespace that matches the pr number.
Here's an [example PR ](https://github.com/SamTheisens/mpyl-example-argocd/pull/3/files#diff-e842e423095b5fe930c3c15ea147f3b91793e47a3d6f717eddf97f0b8e51dd03). The manifests are mapped to application via [this generator](https://github.com/SamTheisens/mpyl-example-argocd/blob/main/argo/pullrequest_generator.yaml)
And this is what its ArgoCD counterpart looks like:
Screenshot 2023-11-04 at 09 45 03

The deployment operations per stage could be done as:
 - _PR_ push a revision to a pull request on the artifacts repo. This PR will never be merged to master. They can be closed periodically to clean up old preview environments.
 - _Acceptance_ build artifacts with MPyL in the usual way and push the artifacts to a branch (with PR for more transparency and convenience) on the artifacts repo. The _merge_ of this PR is what effectively deploys the applications.
 - _Production_ process would be identical to acceptance. Only difference is the target folder in the manifests repo. A rollback would be achieved by: create revert PR, review the manifest diffs, merge it.

I've tried out the PR scenario with `mpyl build artifacts push` https://github.com/SamTheisens/mpyl-example-gha/blob/main/.github/workflows/build-mpyl-pipeline.yml#L61 and this seems to work. Acc and prod could be done the same way. Just a different target folder if I'm not mistaken.

WDYT?

ptab: I actually meant something different with my previous comment, but there's so much great info in your comment that I'll start with that 😄 

> _Production_ The manifest repo is a mirror image of your infrastructure. The manifests in this [folder](https://github.com/SamTheisens/mpyl-example-argocd/tree/main/manifests/projects) are mapped nearly 1-to-1 to applications via [this generator](https://github.com/SamTheisens/mpyl-example-argocd/blob/main/argo/monorepo_application.yaml).
The generator point to HEAD of main. Any revision pushed to main results in a deployment of one or more applications, depending which applications' manifests were altered.
_Acceptance_ should be as much as possible identical to production. The acceptance and production generators cannot point to the same folder with manifests, because the manifests will differ (slightly). You could use separate a folder for test, acceptance and production.

> _Acceptance_ build artifacts with MPyL in the usual way and push the artifacts to a branch (with PR for more transparency and convenience) on the artifacts repo. The merge of this PR is what effectively deploys the applications.
_Production_ process would be identical to acceptance. Only difference is the target folder in the manifests repo. A rollback would be achieved by: create revert PR, review the manifest diffs, merge it.

This is pretty much the setup we've configured so far, down to the usage of the Git generator with a set of folders per environment. Glad to see we're aligned 💪
The only difference is that currently we're committing directly to the `main` branch of the manifests repository (simply because it was faster to build in MPyL), but I 100% agree with you that creating a PR there is a much better option and I want to invest in that before rolling this out company-wide.


> _PR_ for each pull request in your source code repo a counterpart is created in the artifact repo. Only the applications that were altered will receive alterations to the manifests. These manifests are deployed into a namespace that matches the pr number.
_PR_ push a revision to a pull request on the artifacts repo. This PR will never be merged to master. They can be closed periodically to clean up old preview environments.

This is where our setup differs. For simplicity's sake, we're also using a Git directory generator here (like with the "regular" environments), using a folder named `pr-<number>` as the base (instead of `test`, `acceptance` or `production`). When a PR is closed, we have a Github action that removes that directory from the manifests repository. I have to say I really like your approach with the PR generator, and I'll look into that more closely - thanks for the suggestion and example code!

---

Now, to clarify my previous comment: a single code change in our monorepo might span multiple services, and the way MPyL currently works is by committing once per sub-module. So, currently, one single "deployment" might actually consist of multiple commits to the manifests repository. I believe it would be better to group all these changes, either by:
* making them all part of a single commit; or
* making all the commits part of the same branch / pull request

This would IMHO make it much easier to review and rollback deployments.

ptab:
One other thing I noticed: can you also remove [this `managed-by` label](https://github.com/Vandebron/mpyl/blob/dc3455c5d98bfefa3a24e26d93007947a22efcfe/src/mpyl/steps/deploy/k8s/chart.py#L239) from the default manifest created? This label should now be set by Argo during application creation.

src/mpyl/steps/deploy/k8s/kubectl.py
SamTheisens: I think I understand wat you are trying to do. I completely agree, deploys should span all services that were touched in the context of a PR.
But that functionality is _already_ present in the latest version of MPyL :-) Just run `mpyl build artifacts push --pr 1234 --path manifests --artifact-type manifests` at the end of your mpyl run. As demonstrated in this example github pipeline https://github.com/SamTheisens/mpyl-example-gha/blob/main/.github/workflows/build-mpyl-pipeline.yml

SamTheisens:
> One other thing I noticed: can you also remove [this `managed-by` label](https://github.com/Vandebron/mpyl/blob/dc3455c5d98bfefa3a24e26d93007947a22efcfe/src/mpyl/steps/deploy/k8s/chart.py#L239) from the default manifest created? This label should now be set by Argo during application creation.

Indeed, the reference to HELM can be removed. ArgoCD doesn't seem to insert its own labels though. Not sure how it keeps track of its applications.

Jorg88:
Note to self: look into (& fix) whitelists from 1 project being added to another when using `KubectlManifest`

src/mpyl/utilities/repo/__init__.py
```diff
@@ -138,14 +138,17 @@ def from_clone(config: RepoConfig, repo_path: Path):
...
         repo = Repo.clone_from(
-            url=creds.to_url_with_credentials,
+            url=creds.to_url_with_credentials if user_name else creds.ssh_url,
             to_path=repo_path,
         )
-        user_name = creds.user_name
-        with repo.config_writer() as writer:
-            writer.set_value("user", "name", user_name)
-            writer.set_value("user", "email", creds.email or "somebody@somwhere.com")
+        if user_name:
```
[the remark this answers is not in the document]
Jorg88: that would look uglier actually
```
        user_name = creds.user_name
        if user_name:
            repo = Repo.clone_from(
                url=creds.to_url_with_credentials,
                to_path=repo_path,
            )
            with repo.config_writer() as writer:
                writer.set_value("user", "name", user_name)
                writer.set_value(
                    "user", "email", creds.email or "somebody@somwhere.com"
                )
        else:
            Repo.clone_from(
                url=creds.ssh_url,
                to_path=repo_path,
            )
```

src/mpyl/schema/mpyl_config.schema.yml
Jorg88: We decided on `artifactRepository` -> `cachingRepository` & `argoRepository` -> `artifactRepository`

SamTheisens: Why `artifactRepository`? It's not storing any artifacts AFAICS. Rather it stores manifests.

src/mpyl/steps/deploy/k8s/kubectl.py
SamTheisens: Small note on the automated removal of closed/merged PR and their preview environments in the test cluster: this may be quicker than convenient. In practice I've found it convenient for the PR environment to linger a while, so you can (for example) replay a test scenario in the PR env to compare it to what happens in acceptance.
A periodic cleanup of closed PRs older than x days is conceptually less clean, but perhaps more workable in practice.

src/mpyl/schema/mpyl_config.schema.yml
Jorg88: Well we picked kubectl manifests for now, but you could store whatever build artifact you want.

SamTheisens: Indeed, you could also use HELM charts. But in that case would you call these artifacts or manifests? I would call them manifests. Artifacts contain executable code, like .jars or docker containers.

src/mpyl/steps/deploy/k8s/kubectl.py
Jorg88: I think I solved all comments in this thread, please comment on specific pieces of code if I missed something. Easier to track that way!

src/mpyl/steps/deploy/k8s/__init__.py
```diff
@@ -219,13 +184,23 @@ def deploy_helm_chart(  # pylint: disable=too-many-locals
...
+        )
+        path = Path(
+            project.root_path,
+            deployment_config.output_path,
+            "k8s-manifests",
+            project.name,
+            argo_folder_name,
+            namespace,
+        )
+        file_path = write_manifest(path, chart)
```
SamTheisens, line 203: Ik think I would find it cleaner to leave the argo specifics out of here and make that a concern of the `push` logic and the file path mappers. That way the logic here can stay simple and generic.
It also gives more flexibility to the end-user to choose different path structures if they'd want to.

SamTheisens (approved):
It looks nice to me! Just one remark (see comment).

FYI, I have renamed the `artifactRepository` to `argoRepository`. It's more consistent (with `argoGIthub`) and the word `manifest` is needlessly generic I think. At this point ArgoCD is its only usecase and I don't expect that to change any time soon.

src/mpyl/steps/deploy/k8s/__init__.py
Jorg88: I tried to implement this, and got quite far.. but I don't see a way to get the project name (the name from the `project.yml`, not from the project path) outside of the actual step. Maybe you or Daniel see a way, feel free to comment or implement in that case.

SamTheisens: Why does the name of the project path not suffice? You _could_ load the project in the path transformer I guess?

PR #314, opened 2023-11-16
Bug/data 1165 fix project schema env class

kgrunert:


Jorg88:
@kgrunert this pr seems to be stale? do you still want to pick it up or should we close it?

kgrunert:
> @kgrunert this pr seems to be stale? do you still want to pick it up or should we close it?

We can close it! We thought at first it was a bug, but it was valid. no clue why i havent followed up on that

PR #317, opened 2023-11-29
Set `DOCKER_IMAGE` with full path

koeves:
----
 # ⚠️ Could not find ticket corresponding to `feature/fix_docker_image_tag. Does your branch name follow the correct pattern?

### DIFF
--- a/src/mpyl/steps/test/dockertest.py
+++ b/src/mpyl/steps/test/dockertest.py
@@ -40,6 +40,7 @@ from ...utilities.docker import (
     push_to_registry,
     registry_for_project,
     get_default_build_args,
+    full_image_path_for_project,
 )
 from ...utilities.junit import (
     to_test_suites,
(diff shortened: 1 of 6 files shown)


src/mpyl/steps/test/dockertest.py
```diff
@@ -82,7 +83,7 @@ def execute(self, step_input: Input) -> Output:
             target=test_target,
             registry_config=docker_registry_config,
             build_args=get_default_build_args(
-                tag,
+                full_image_path_for_project(step_input),
```
Jorg88, line 86: Like I said on slack: I think the `if step_input.dry_run` should move into the function to get the path to prevent the copied logic. And I don't understand why we don't need to take `dry_run` into account for the docker steps? 

I guess those 2 questions/comments very much relate 😛

koeves: [0faca25](https://github.com/Vandebron/mpyl/pull/317/commits/0faca2538986c9b05d0f83f6fca741dadef1dc24)

Jorg88 (approved):

PR #318, opened 2023-11-29
Feature/tech 701 parallel runs

Jorg88:


src/mpyl/schema/run_properties.schema.yml
```diff
@@ -108,3 +108,6 @@ properties:
           type: string
         icon:
           type: string
+        parallelism_factor:
```
SamTheisens, line 111: It's a nit, but the word `parallelism` would be enough. In the description you could explain further.

src/mpyl/utilities/parallel/__init__.py
```diff
@@ -0,0 +1,35 @@
+"""Utility tool for running commands in parallel"""
```
SamTheisens, line 1: Utility and tool is double. I would prefer just Utility.

src/mpyl/utilities/parallel/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+from concurrent.futures import Future, ThreadPoolExecutor
+from dataclasses import dataclass
+from typing import Callable, Any, TypeVar, Iterable
+
+T = TypeVar("T")
+
+
+@dataclass(frozen=True)
+class ParallelObject:
```
SamTheisens, line 10: Perhaps `ParallelCommand` would be more specific?

src/mpyl/utilities/parallel/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+    function: Callable
+    parameters: dict[Any, Any]
```
SamTheisens, line 12: I guess the keys in this dict are always strings?

src/mpyl/utilities/sbt/__init__.py
```diff
@@ -35,7 +35,11 @@ def from_config(config: dict):
         )
 
     def to_command(self, client_mode: bool, sbt_commands: list[str]):
-        cmd = [self.sbt_client_command if client_mode else self.sbt_command]
+        cmd = (
+            self.sbt_client_command.split(" ")
```
SamTheisens, line 39: I guess you can do the split once outside the braces?

src/mpyl/utilities/subprocess/__init__.py
```diff
@@ -21,6 +21,8 @@ def custom_check_output(
 
     command_argument = " ".join(command)
     logger.info(f"Executing: '{command_argument}'")
+    logger = logger.getChild("Subprocess")
```
SamTheisens, line 24: What does this do to what is printed in the logs? Will they they be interleaved?

src/mpyl/build.py
```diff
@@ -239,17 +239,32 @@ def run_build(
...
+                    function=executor.execute,
+                    parameters={
+                        "stage": stage.name,
+                        "project": project,
+                        "dry_run": dry_run,
+                    },
+                )
+                for project in projects
+            ]
+            results: list[StepResult] = run_in_parallel(
```
SamTheisens, line 253: So if I understand correctly, we fail per _stage_. If one of the steps fails immediately, the others are allowed to complete?

SamTheisens (commented):
Looks cool! It's fortunately not very invasive. Main question I have is what it does to the logging. Interleaving is what would naturally occur, but that's hard to read.
You could also do something along these lines https://github.com/Vandebron/mpyl/blob/main/validate.py#L74-L85
where pile up the logs and only dump them at the end for the task that has failed. For local build you could show a spinner for the tasks that are running.

The logging experience that you get with dagster is unsurpassed: filtering by task and log level after the fact. I wonder if you could publish the logs in a binary format that can be interpreted by the dagster GUI or something like that. More research is needed...

src/mpyl/utilities/parallel/__init__.py
Jorg88: Yes indeed, there was a point where they weren't and I forgot the change it back.

src/mpyl/utilities/subprocess/__init__.py
Jorg88: It doesn't seem to interleave it any more/less than with a print, at least it uses the logger now (with the nice formatting etc) and because of the child name it's a bit more clear what is happening.

The only downside is that is prints an extra info `INFO     mpyl.Subprocess  \[info]` that I can't seem to get rid off 😓

src/mpyl/build.py
SamTheisens: I think that strategy indeed makes the most sense. We could also terminate the other threads as soon as the first one fails, but we would need to introduce another "terminated" fail state. 
And although your build might run a bit longer than strictly necessary, you will get (probably) cacheable outcomes. And you also get potentially useful information to find out what's wrong from multiple failed build outputs.

So yes, failing on stage level makes most sense to me too.

Jorg88:
This pr is on hold until we or the backend guild finds a solution for sbt commands and parallel threads.

PR #320, opened 2023-12-05
feature/TECH 713 fix manual selection deploy

Jorg88:
- [TECH-713] Move the build plan into the run properties
- [TECH-713] Fix find_deploy_set for manual selection runs
- [TECH-713] Replace find_deploy_set with re-used logic in RunProperties

[TECH-713]: https://vandebron.atlassian.net/browse/TECH-713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TECH-713]: https://vandebron.atlassian.net/browse/TECH-713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TECH-713]: https://vandebron.atlassian.net/browse/TECH-713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

----
 # ⚠️ Could not find jira ticket with id TECH-731

### DIFF
--- a/mpyl-example.py
+++ b/mpyl-example.py
@@ -2,27 +2,22 @@ import argparse
 import logging
 import sys
 from logging import Logger
-from pathlib import Path
 
 from rich.console import Console
 from rich.logging import RichHandler
-
-from mpyl import Repository, RepoConfig
-from mpyl.project import Stage, load_project
-from mpyl.stages.discovery import find_build_set
 
 
 def main(log: Logger, args: argparse.Namespace):
     if args.local:
         from src.mpyl.reporting.targets.jira import JiraReporter
-        from src.mpyl.steps.models import RunProperties
+        from src.mpyl.steps.run_properties import initiate_run_properties
         from src.mpyl.utilities.pyaml_env import parse_config
         from src.mpyl.cli import MpylCliParameters
         from mpyl.build import run_mpyl
 
     else:
         from mpyl.reporting.targets.jira import JiraReporter
-        from mpyl.steps.models import RunProperties
+        from mpyl.steps.run_properties import initiate_run_properties
         from mpyl.utilities.pyaml_env import parse_config
         from mpyl.build import run_mpyl
         from mpyl.cli import MpylCliParameters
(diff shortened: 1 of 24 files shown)


mpyl-example.py
```diff
@@ -24,9 +29,51 @@ def main(log: Logger, args: argparse.Namespace):
...
-    run_properties = RunProperties.from_configuration(
-        run_properties=properties, config=config, cli_tag=args.tag
+    cli_parameters = MpylCliParameters(
+        local=args.local,
+        tag=args.tag,
+        pull_main=True,
+        verbose=args.verbose,
+        all=args.all,
     )
+    with Repository(RepoConfig.from_config(config)) as repo:
```
koeves, line 39: Is it possible to centralise this logic of constructing the RunProperties? Because now it's sort of repeated here and also in src/mpyl/build.py and also in src/mpyl/cli/build.py

src/mpyl/steps/models.py
```diff
@@ -92,35 +93,41 @@ class RunProperties:
...
     @staticmethod
     def for_local_run(
         config: dict,
+        run_plan: dict[Stage, set[Project]],
         revision: str,
         branch: Optional[str],
-        tag: Optional[str],
-        stages: list[dict],
+        tag: Optional[str] = None,
+        stages: Optional[list[dict]] = None,
```
koeves, line 109: Why is this optional? Also what's the difference between None and [] in this case?

src/mpyl/steps/models.py
```diff
@@ -92,35 +93,41 @@ class RunProperties:
...
-        if stage_by_name:
-            return stage_by_name
-        raise ValueError(f"Stage {stage_name} not found")
-
     @staticmethod
     def from_configuration(
         run_properties: dict,
         config: dict,
+        run_plan: Optional[dict[Stage, set[Project]]] = None,
+        all_projects: Optional[set[Project]] = None,
```
koeves, line 130: Same here. If you make it non optional and pass an empty dict or set as params, you wouldn't need the `run_plan or {}` part below

koeves (changes requested):
Looks good in general, just small things

src/mpyl/steps/run_properties.py
```diff
@@ -0,0 +1,77 @@
...
+                        )
+                    ),
+                    stages=stages,
+                    build_all=cli_parameters.all,
+                    selected_stage=cli_parameters.stage,
+                    selected_projects=cli_parameters.projects,
+                )
+
+    if cli_parameters.local:
+        return RunProperties.for_local_run(
```
koeves, line 61: Do we still need this?

Jorg88: We're definitely closer to not needing it anymore, but I think there is still some extra work if you want to get rid of it. Maybe we can make a ticket? I think we should get this bugfix out and I think the current solution is pretty clean already since the function is only called here and there isn't a lot of extra logic around it anymore.

koeves (approved):

src/mpyl/steps/run_properties.py
```diff
@@ -0,0 +1,77 @@
...
+    all_projects: Optional[set[Project]] = None,
+) -> RunProperties:
+    if all_projects is None or run_plan is None:
+        with Repository(RepoConfig.from_config(config)) as repo:
+            if all_projects is None:
+                project_paths = repo.find_projects()
+                all_projects = set(
+                    map(
+                        lambda p: load_project(
+                            root_dir=Path(""),
```
SamTheisens, line 27: The root dir needs to be passed in from the outside, otherwise it's a pointless parameter.

PR #335, opened 2024-01-04
feature/selectable target clusters

koeves:
## Selectable clusters

1. Be able to specify a list of clusters to deploy to

https://github.com/Vandebron/mpyl/blob/eef58684a221d3182c9c48cc628ca47b4d2e79a6/mpyl_config.example.yml#L110-L130

which can be overridden on a per-project basis:

https://github.com/Vandebron/mpyl/blob/a2803e9a3294d3b46ce8deead277a976213750c1/tests/test_resources/test_project.yml#L24-L29

2. Make default traefik config aware of the cluster the project at hand is deployed to:

Interpolate the cluster's `clusterEnv` field when constructing the ingress route:

https://github.com/Vandebron/mpyl/blob/a2803e9a3294d3b46ce8deead277a976213750c1/mpyl_config.example.yml#L150-L157

---
## Follow-up functionality for IngressRoutes:

Be able to specify an additional IngressRoute per host in using different middleware or entrypoint
https://github.com/Vandebron/mpyl/blob/d04df2b52a5dd68e9a05b39ff065d979fd6e01b0/tests/test_resources/mpyl_config.yml#L103-L108

which can be referred to in the project yaml: https://github.com/Vandebron/mpyl/blob/8f9fe85a9abbd0f1f9e6c27bd024b96e088f9fa5/tests/test_resources/test_project.yml#L130


mpyl_config.example.yml
```diff
@@ -82,23 +82,27 @@ whiteLists:
...
-        clusterId: c-acme
-        projectId: p-acme
-        clusterEnv: acce
-        context: acme-k8s-acce
-      production:
-        clusterId: c-acme
-        projectId: p-lb52t
-        clusterEnv: prd
-        context: acme-k8s-prod
+  mainCluster:
```
SamTheisens, line 85: maybe this is more like the `defaultCluster`?

mpyl_config.example.yml
```diff
@@ -147,10 +151,10 @@ project: # default values
       enabled: true
       hosts:
         - host:
-            pr: "Host(`{SERVICE-NAME}-{PR-NUMBER}.test-backend.nl`)"
-            test: "Host(`{namespace}-{SERVICE-NAME}.test-backend.nl`)"
-            acceptance: "Host(`{namespace}-{SERVICE-NAME}.acce-backend.nl`)"
-            production: "Host(`{namespace}-{SERVICE-NAME}.prod-backend.nl`)"
+            pr: "Host(`{SERVICE-NAME}-{PR-NUMBER}.{CLUSTER-ENV}-backend.nl`)"
```
SamTheisens, line 154: I think it would be good to document how the substitution for different targets works and what the possible wildcards are that you can use.

SamTheisens (commented):
Looks nice 👍 . What about argoCD? Would you push to different repos in that case? 

Two comments:
 - I don't see a roundtrip test to validate that namespace metadata ends up as a expected.
 - Some documentation about the PR preview env mechanism and what variables can be substituted.

mpyl_config.example.yml
```diff
@@ -169,12 +176,23 @@ project: # default values
...
+            production: "Host(`{namespace}-{SERVICE-NAME}.{CLUSTER-ENV}-backend.nl`)"
           tls:
             all: "le-custom-prod-wildcard-cert"
           insecure: false
           whitelists:
             all: [ "VPN" ]
+    additionalTraefikRoutes:
+      - name: "ingress-intracloud-https"
+        clusterEnv:
+          pr: "test-other"
```
Jorg88, line 191:
```suggestion
          pr: "pr-other"
```

mpyl_config.example.yml
```diff
@@ -169,12 +176,23 @@ project: # default values
...
+          test: "test-other"
+          acceptance: "acce-other"
+          production: "acce-other"
```
Jorg88, line 194:
```suggestion
          production: "prod-other"
```

releases/notes/1.6.7.md
```diff
@@ -0,0 +1,52 @@
...
+
+- Be able to specify a list of target clusters to deploy to
+- Be able to specify default targets per environment
+- Be able to override the default targets in the `project.yaml`
+
+```yaml
+deployment:
+  cluster:
+    pr: 'other-test'
+    test: 'test'
```
Jorg88, line 11: Maybe use k8s & leaf as example to make the point clear

src/mpyl/projects/versioning.py
```diff
@@ -113,8 +113,8 @@ def upgrade(self, previous_dict: ordereddict) -> ordereddict:
         return previous_dict
 
 
-class ProjectUpgraderOneFour20(Upgrader):
-    target_version = "1.4.20"
+class ProjectUpgraderOneSix7(Upgrader):
```
Jorg88, line 116: I'm confused, why are you replacing the 4.20 upgrader instead of adding a new one?

src/mpyl/projects/versioning.py
```diff
@@ -203,12 +203,42 @@ def upgrade(self, previous_dict: ordereddict) -> ordereddict:
...
+                    kubernetes.pop("rancher")
+
+                    kubernetes["defaultCluster"] = {
+                        env: next(
+                            (
+                                c["clusterEnv"]
+                                for c in named_clusters
+                                if c["name"] == env
+                            ),
+                            "MISSING",
```
Jorg88, line 232: What idea did you have with this default value?

src/mpyl/schema/mpyl_config.schema.yml
```diff
@@ -134,24 +138,48 @@ definitions:
         description: >-
           The path to the folder where the rendered helm chart or kubectl manifest will be written.
           Defaults to target/kubernetes under the root folder of the project.
-      rancher:
+      defaultCluster:
         type: object
-        additionalProperties: false
+        description: Main cluster to deploy to. Can be overriden in the project.yaml
```
Jorg88, line 143:
```suggestion
        description: Default cluster to deploy to. Can be overriden in the project.yaml
```

src/mpyl/steps/deploy/k8s/__init__.py
```diff
@@ -90,17 +90,15 @@ def upsert_namespace(
...
     logger.info(
-        f"Deploying target {run_properties.target} and k8s context {rancher_config.context}"
+        f"Deploying target {run_properties.target} and k8s context {cluster_config.context}"
     )
     api = client.CoreV1Api()
 
-    meta_data = rancher_namespace_metadata(
-        namespace=namespace, rancher_config=rancher_config, project_id=project_id
-    )
+    meta_data = get_namespace_metadata(namespace, cluster_config, project_id)
```
Jorg88, line 101: I thought we would try to keep & add the `x=x` syntax for function params? 😬

src/mpyl/steps/deploy/k8s/cluster.py
```diff
@@ -0,0 +1,81 @@
...
+    )
+
+    if not default_cluster:
+        raise ValueError(
+            f"Default cluster {default_cluster_name} not found in list of clusters"
+        )
+
+    cluster_for_env = next(
+        (cluster for cluster in clusters if cluster.name == cluster_override),
+        default_cluster,
```
Jorg88, line 58: Shouldn't we throw an error instead of using the default if an override is defined but it doesn't exist?

src/mpyl/steps/deploy/k8s/cluster.py
```diff
@@ -0,0 +1,81 @@
...
+    )
+
+    return cluster_for_env
+
+
+def get_namespace_metadata(
+    namespace: str, cluster_config: ClusterConfig, project_id: str
+):
+    metadata: dict[str, Any] = {
```
Jorg88, line 67: Can't you use `dict` instead of `Any`?

koeves: it's more Union[str, dict].. I'm just going to remove the type hint, it's not very useful here

koeves: actually type check fails then, so I'll leave it with `Any`

src/mpyl/projects/versioning.py
koeves: that we just write MISSING so it's obviously not right

src/mpyl/steps/deploy/k8s/cluster.py
koeves: hm I guess we could, but how do you differentiate a cluster override not being found vs not being defined?

src/mpyl/project.py
```diff
@@ -340,6 +340,23 @@ def from_config(values: dict):
...
 
+@dataclass(frozen=True)
+class TraefikAdditionalRoute:
+    name: str
+    middlewares: list[str]
+    entrypoints: list[str]
+
+    @staticmethod
+    def from_config(values: dict):
+        if not values:
```
Jorg88, line 351: shouldn't `values` be an optional param then?

mpyl_config.example.yml
```diff
@@ -169,12 +176,23 @@ project: # default values
...
+        middlewares:
+          - "intracloud-middleware@kubernetescrd"
+        entrypoints:
+          - "intracloud"
```
Jorg88, line 198: Random note: I think I've seen @ptab mention somewhere that having a newline at the end of your files should be the default

src/mpyl/schema/mpyl_config.schema.yml
```diff
@@ -134,24 +138,48 @@ definitions:
...
-        "$ref": "#/definitions/RancherCluster"
+        "$ref": "#/definitions/Cluster"
       acceptance:
-        "$ref": "#/definitions/RancherCluster"
+        "$ref": "#/definitions/Cluster"
       production:
-        "$ref": "#/definitions/RancherCluster"
-  RancherCluster:
+        "$ref": "#/definitions/Cluster"
+  NamedCluster:
```
Jorg88, line 165: Am I missing something or is this change not reflected in the example config?

src/mpyl/steps/deploy/k8s/chart.py
```diff
@@ -509,12 +518,25 @@ def to_white_list(
                 white_lists=to_white_list(host.whitelists),
                 tls=host.tls.get_value(self.target) if host.tls else None,
                 insecure=host.insecure,
+                additional_route=next(
+                    (
+                        r
+                        for r in self.config_defaults.additional_routes
+                        if r.name == host.additional_route
```
Jorg88, line 525:
```suggestion
                        route
                        for route in self.config_defaults.additional_routes
                        if route.name == host.additional_route
```

src/mpyl/steps/deploy/k8s/cluster.py
```diff
@@ -0,0 +1,85 @@
...
+            cluster_id=config.get("clusterId"),
+            cluster_env=config["clusterEnv"],
+            context=config["context"],
+        )
+
+
+def get_cluster_config_for_project(
+    run_properties: RunProperties, project: Project
+) -> ClusterConfig:
+    cluster_override_name = (
```
Jorg88, line 30: maybe move this to just before line 56?

src/mpyl/steps/deploy/k8s/cluster.py
```diff
@@ -0,0 +1,85 @@
...
+    clusters = [
+        ClusterConfig.from_config(cluster_config)
+        for cluster_config in kubernetes_config["clusters"]
+    ]
+
+    default_cluster_name: str = TargetProperty.from_config(
+        kubernetes_config["defaultCluster"]
+    ).get_value(run_properties.target)
+
+    default_cluster: ClusterConfig = next(
```
Jorg88, line 47: I would maybe extract the logic to get the default cluster to its own function and only call it in the `else` on line 65. Makes this whole thing easier to read about what's happening

src/mpyl/steps/deploy/k8s/resources/traefik.py
```diff
@@ -27,34 +28,45 @@ def full_name(self) -> str:
...
+        def _interpolate_names(host: str, name: str, cluster_env: str) -> str:
             host = host.replace("{SERVICE-NAME}", name)
             host = host.replace("{namespace}", namespace)
+            host = host.replace("{CLUSTER-ENV}", cluster_env)
             host = replace_pr_number(host, pr_number)
             return host
 
+        combined_middlewares = (
+            [
+                {"name": "traefik-https-redirect@kubernetescrd"} if not https else None,
```
Jorg88, line 52: I know this was already hardcoded, but should it be?

src/mpyl/steps/deploy/k8s/resources/traefik.py
```diff
@@ -66,13 +78,19 @@ def _interpolate_names(host: str, name: str) -> str:
         if host.insecure:
             tls |= {"options": {"name": "insecure-ciphers", "namespace": "traefik"}}
 
+        combined_entrypoints = (
+            ["websecure" if https else "web"]
```
Jorg88, line 82: same here

src/mpyl/schema/mpyl_config.schema.yml
koeves: what do you mean?

src/mpyl/steps/deploy/k8s/resources/traefik.py
koeves: no it probably shouldn't be. Do you want me to change it here?

src/mpyl/schema/mpyl_config.schema.yml
Jorg88: Nvm, I was looking at it wrong!

src/mpyl/projects/versioning.py
Jorg88: I was thinking that it's a bit obscure maybe, why not throw an error instead? Because this will error out later anyway right?

src/mpyl/steps/deploy/k8s/resources/traefik.py
Jorg88: Normally I would say you could create a new ticket, but this kinda fits within the scope of this pr right?

src/mpyl/steps/deploy/k8s/cluster.py
koeves: done in [d297458](https://github.com/Vandebron/mpyl/pull/335/commits/d2974582b15707f23620907591b70514a1524a36)

moved it around, but didn't do exactly like you said, so we can still keep the exception if the default is not found

src/mpyl/steps/deploy/k8s/resources/traefik.py
koeves: these are actually just traefik constants: https://doc.traefik.io/traefik/routing/entrypoints/

src/mpyl/steps/deploy/k8s/resources/traefik.py
koeves: done in [e4b7a9b](https://github.com/Vandebron/mpyl/pull/335/commits/e4b7a9be824e2c3b9d273d14fc5bf0a447a2ed93)

Jorg88 (approved):
Nice work Daniel!

PR #340, opened 2024-01-17
[TECH-520] SealedSecrets for dagster deployment

kgrunert:
**I think it is best to view this PR as a whole and not commit by commit as I have been trial-and-error'ring a bit here**

With this PR I would like to introduce the integration of Sealed Secrets in dagster!
Luckily dagster's user-deployment helm chart made that easy by allowing for `extraManifests` to be deployed along its default resources.

That way I could use the `ChartBuilder` to build our custom resource SealedSecret and convert it to a dict and attach it to the user-code `values.yaml` values

There is a difference to the conventional Sealed Secret regarding naming and namespace-ing, due to the nature how dagster user-deployments are being deployed:
1. PRs are deployed as of now into the same namespace (`dagster`) so the release-name for PRs has the `-pr-<TAG>` as a suffix and the same needs to apply for the SealedSecret resource, which is done in the to_user_code_values function by overwriting the release name of that SealedSecret resource

It is a bit of a shame that we can't use the `ChartBuilder` (or maybe we can and i am overseeing it let me know!) for the DagsterDeploy as it is waaaay easier (and more maintainable) to just write your `values.yaml` for this helm chart rather than writing all charts yourself.

EDIT:
> 1. the chartbuilder has the wonderful function of `_create_sealed_secret_env_vars` which i had to duplicate for the dagsterdeploy step, the alternative i also tried out but discarded was to handcraft the `dict` 

i see the flaw right now in here and would need to make some adjustments, it contains some none values when doing the `to_dict()`-transformation :/ , those `None`s need to go, maybe handcrafting this is not a bad idea after all...
```
{'name': 'test', 'value': None, 'valueFrom': {'configMapKeyRef': None, 'fieldRef': None, 'resourceFieldRef': None, 'secretKeyRef': {'key': 'test', 'name': 'example-dagster-user-code-pr-1234', 'optional': False}}}
```

Attempted to fix it here: [ba3623f](https://github.com/Vandebron/mpyl/pull/340/commits/ba3623f96d8c59e5de06e5915c177f07ce67f7f0)

Furthermore, the to_user_code_values is now relying a bit more on the ChartBuilder, so that we dont have to reinvent the wheel when writing SealedSecrets and env-vars

----
### 📕 [TECH-520](https://vandebron.atlassian.net/browse/TECH-520) As CI/CD I would like to inject sealed secrets into Dagster User Code Repositories pietercusters@vandebron.nl 
Currently, secrets are being set as `envSecret`s in the helm chart for a dagster user code repository. 

[edit#slide=id.g198082d3ba3*0_81](https://docs.google.com/presentation/d/1gCAFtZuIB9QJ9LHnaqLtdCLdhZqir0FIcZzSlnEgutk/edit#slide=id.g198082d3ba3_0*81)  give also a starting point for dagster’s user code

Goal of this card is to explore the possibility of using sealed secrets when doing a Dagster Deploy with MPyL, `Kubernetes Deploy}} and {{Kubernetes Job Deploy`, already support this functionality.

🏗️ Build [9](https://jenkins.k8s-serv-backend.vdbinfra.nl/job/MPyL%20Pipeline%20-%20Test/job/PR-340/9/display/redirect) ✅ Successful, started by _Katrin Grunert_  
🚀 *[cloudfront-service](https://cloudfront-service-340.test.nl/swagger/index.html)*, *example-dagster-user-code*, *job*, *kong-sync*, *[nodeservice](https://nodeservice-340.test.nl/swagger/index.html)*, *ocpp-first*, *ocpp-second*, *[sbtservice](https://sbtservice-340.test.nl/swagger/index.html)*, *sparkJob-first*, *sparkJob-second*  

[TECH-520]: https://vandebron.atlassian.net/browse/TECH-520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/mpyl/steps/deploy/k8s/chart.py
```diff
@@ -593,12 +593,14 @@ def to_role_binding(self) -> V1RoleBinding:
...
 
-    def to_sealed_secrets(self) -> V1SealedSecret:
+    def to_sealed_secrets(self, release_name: Optional[str] = None) -> V1SealedSecret:
         secrets: dict[str, str] = {}
         for secret in self.sealed_secrets:
             secrets[secret.key] = secret.get_value(self.target)
 
-        return V1SealedSecret(name=self.release_name, secrets=secrets)
+        return V1SealedSecret(
+            name=release_name if release_name else self.release_name, secrets=secrets
```
Jorg88, line 602: this seems a bit awkward to me, that the `release_name` class prop is not the actual release name?

src/mpyl/steps/deploy/k8s/resources/dagster.py
```diff
@@ -45,14 +63,17 @@ def to_user_code_values(
...
-                "name": f"{project.name}{name_suffix}",
+                "name": release_name,
                 "port": 3030,
                 "resources": {
                     "requests": {"memory": "256Mi", "cpu": "50m"},
                     "limits": {"memory": "512Mi", "cpu": "1000m"},
                 },
             },
         ],
+        "extraManifests": [to_dict(manifest) for manifest in extra_manifests]
```
Jorg88, line 74: What would be the best practice here @koeves , to add an empty list as value or no key/value at all if there are no extra manifests?

src/mpyl/steps/deploy/k8s/chart.py
kgrunert: i agree, alternatively i could alter the release name by accessing `sealed_secret.metadata.name` in the `DeployDagster`-step.

src/mpyl/steps/deploy/dagster.py
```diff
@@ -94,19 +96,51 @@ def execute(self, step_input: Input) -> Output:
...
+            V1EnvVar(
+                name=e.key,
+                value_from=V1EnvVarSource(
+                    secret_key_ref=V1SecretKeySelector(
+                        key=e.key, name=release_name, optional=False
+                    )
+                ),
+            )
+            for e in sealed_secrets_for_target
+        ]
```
koeves, line 126: I would potentially move most of this logic into the `ChartBuilder` (so we localize the k8s client related objects there)

src/mpyl/steps/deploy/k8s/resources/dagster.py
```diff
@@ -18,10 +22,13 @@ class Constants:
 
 def to_user_code_values(
     project: Project,
+    release_name: str,
     name_suffix: str,
     run_properties: RunProperties,
     service_account_override: Optional[str],
     docker_config: DockerConfig,
+    sealed_secrets: Optional[List[V1EnvVar]] = None,
+    extra_manifests: Optional[List[CustomResourceDefinition]] = None,
```
koeves, line 31: Shouldn't these just be `Lists` (without a default argument)? Then you don't need the `if sealed_secrets else []` later on. I think you pass `sealed_secrets_as_env_var` here anyway which is not optional. Maybe I missed something though

src/mpyl/steps/deploy/k8s/resources/dagster.py
koeves: I guess it would look nicer if we don't write `extraManifests` there if the list is empty.. But then, no one is actually looking at this result yaml right? It's just for helm.

koeves (commented):
Nice improvements @kgrunert! 🎉 
Just small comments

src/mpyl/steps/deploy/k8s/resources/dagster.py
kgrunert: no, it's a good comment! i will implement that

src/mpyl/steps/deploy/k8s/resources/dagster.py
kgrunert: yah helm doesnt care whether `extraManifests` exist or not.

src/mpyl/steps/deploy/k8s/chart.py
kgrunert: and for context, we havent setup namespace independent config for dagster user-code yet. thats why we deploy user-code into a singular namespace where everything dagster-related (database,daemon) is also deployed in

src/mpyl/steps/deploy/k8s/resources/sealed_secret.py
```diff
@@ -9,6 +9,7 @@
 
 class V1SealedSecret(CustomResourceDefinition):
     def __init__(self, name: str, secrets: dict[str, str]):
+        self.secrets = secrets
```
Jorg88, line 12: Am I missing something or is this not doing anything, the value also doesn't seem to exist in the (super)class?

kgrunert: hmm so the reason was that i dont necessarily want to deploy a sealedSecret that has no secrets inside. maybe there is a better way to check for this than to call `.secrets`  ?

kgrunert: for example by checking `len(sealed_secret_refs)` ?

kgrunert: did so in https://github.com/Vandebron/mpyl/pull/340/commits/3ffcb53463d845f5fc0f6bd056ab6e74b7db620c

releases/README.md
```diff
@@ -6,6 +6,7 @@
 #### Ship typing information
 - MPyL can now be type checked via `mypy`.
 
+
```
Jorg88, line 9: This file is automatically generated when we create a release, so no need to change it here

Jorg88 (approved):

koeves (approved):
Looks good! The changes to the main releases README can be reverted as Jorg said

PR #343, opened 2024-01-19
update endpoint for nginx

lauraoneasca:


src/mpyl/steps/deploy/kubernetes.py
```diff
@@ -66,7 +66,12 @@ def execute(self, step_input: Input) -> Output:
                     f"Service {step_input.project.name} reachable at: {hostname}"
                 )
                 endpoint = (
-                    "/" if has_specific_routes_configured else "/swagger/index.html"
+                    "/"
+                    if (
+                        has_specific_routes_configured
+                        or "nginx" in step_input.project.name
```
Jorg88, line 72: Did you purposely use a 'contains' check or did you want to use an 'equals' (`==`)? Looks good otherwise!

Jorg88 (approved):

src/mpyl/project.py
```diff
@@ -343,6 +343,7 @@ def from_config(values: dict):
 class TraefikHost:
     host: TargetProperty[str]
     service_port: Optional[int]
+    has_swagger: bool
```
Jorg88, line 346: this can be optional right?

koeves (commented):
Nice first PR 🎉 

You can add a release note in `releases/notes`

koeves (approved):

src/mpyl/steps/deploy/kubernetes.py
```diff
@@ -46,6 +46,28 @@ def try_extract_hostname(
                 return DeployKubernetes.match_to_url(match)
         return None
 
+    @staticmethod
+    def get_endpoint(builder: ChartBuilder) -> str:
+        step_input = builder.step_input
+        has_specific_routes_configured: bool = bool(
```
Jorg88, line 52: Did you guys get errors, or why add the explicit type & typecasting? It's not needed afaik, the inferred type is already `bool`

src/mpyl/steps/deploy/kubernetes.py
```diff
@@ -46,6 +46,28 @@ def try_extract_hostname(
...
+        )
+        hosts = (
+            step_input.project.deployment.traefik.hosts
+            if step_input.project.deployment and step_input.project.deployment.traefik
+            else []
+        )
+        if hosts:
+            has_swagger = hosts[0].has_swagger
+        else:
+            has_swagger = True
```
Jorg88, line 64:
```suggestion
        has_swagger = hosts[0].has_swagger if hosts else True
```

Jorg88 (commented):
Good to see a new contributor, nice work!

src/mpyl/steps/deploy/kubernetes.py
lauraoneasca: that code was already there, I only moved it

The type is inferred indeed, removed that 🧹 if I remove the typecasting as well, the inferred type becomes `Any | Literal[False]`. Unclear if it's something related to my setup, because running `pipenv run validate` seems fine locally 🤔

Jorg88: Right of course, sorry! 

I tried it myself and for me it does just show that the type is `bool`. I think we also do this in more places, but it's hard to search for it. I can't screenshot the inferred type mouseover, but here's the normal one:

Screenshot 2024-01-23 at 21 01 00

lauraoneasca: hmm then it must be something on my side. Updated, removed this typecasting 🧹

Jorg88 (approved):

PR #349, opened 2024-01-29
Check if ECR repo exists

ewaldschrader:
# Ticket:  [Automate ECR repository creation](https://vandebron.atlassian.net/browse/DOP-1203?atlOrigin=eyJpIjoiOWE5MTM1OTI2MGY1NDgyZmE0MzNkYjliMGI2MWZkN2YiLCJwIjoiaiJ9)

## TEST REPORT:
The function `create_ecr_repo_if_needed` has been created to check if the image is pushed to an existing ECR repository. It has been tested in the following ways: 
* **Pushing an image with an existing repository:** 
The image is pushed to the existing repository with the given tag. No further action is needed.
* **Pushing an image with a non-existing repository name:**
Throws the exception `RepositoryNotFoundException`, this causes a new repository to be created and the function `ecr_lifecycle_policy` adds a lifecycle policy to the repository. Then the image gets pushed to the repository with its tag.  
* **An new project named `1-job` is pushed to the registry:** 
Throws an exception, as the name convection is not accepted by AWS. 

The test function `test_ecr_repository_path` checks if `ecr_repository_path` returns the correct repository name. 

----

 # ⚠️ Could not find ticket corresponding to `feature/ECR_REPO_CHECK. Does your branch name follow the correct pattern?


src/mpyl/utilities/docker/__init__.py
```diff
@@ -338,3 +340,21 @@ def remove_container(logger: Logger, container: Container) -> None:
...
     docker.remove(container.id)
     logger.info(f"Removed container {container.id}")
+
+
+def check_ecr_repo(logger: Logger, repo):
+    # Initialize the ECR client
+    ecr_client = boto3.client('ecr')
+
+    # Check if the repository already exists
+    try:
```
Jorg88, line 350: Instead of a try catch structure, don't pass any params to the `describe_repositories()` call. According to the documentation that will list all repo's, so then you can search through the list and handle accordingly. Much cleaner piece of code that way!

src/mpyl/utilities/docker/__init__.py
```diff
@@ -338,3 +340,21 @@ def remove_container(logger: Logger, container: Container) -> None:
     logger.debug(f"Removing container {container.id}")
     docker.remove(container.id)
     logger.info(f"Removed container {container.id}")
+
+
+def check_ecr_repo(logger: Logger, repo):
```
Jorg88, line 345:
```suggestion
def create_ecr_repo_if_needed(logger: Logger, repo):
```

src/mpyl/utilities/docker/__init__.py
ewaldschrader: I can do that as well, but doesn't `ecr_client.describe_repositories(repositoryNames=[repo,])` already does this? I don't see how it is cleaner.

src/mpyl/utilities/docker/__init__.py
```diff
@@ -338,3 +340,21 @@ def remove_container(logger: Logger, container: Container) -> None:
     logger.debug(f"Removing container {container.id}")
     docker.remove(container.id)
     logger.info(f"Removed container {container.id}")
+
+
+def create_ecr_repo_if_needed(logger: Logger, repo):
+    # Initialize the ECR client
```
koeves, line 346: I think you can remove the comments in the method, it's clear what the statements are doing

src/mpyl/utilities/docker/__init__.py
koeves: I think we can keep the try-catch, otherwise we need to implement pagination as well (according to their documentation it returns 100 repos at once normally). You can set the max value to a 1000 but then the call would take longer.. Having a small try-catch block with a specific error type should be fine in my opinion.

src/mpyl/utilities/docker/__init__.py
```diff
@@ -190,6 +191,7 @@ def push_to_registry(
 
     login(logger=logger, registry_config=docker_config)
     full_image_path = docker_registry_path(docker_config, image_name)
+    check_ecr_repo(logger=logger, repo=image_name)
```
koeves, line 194: Where is this method defined? Shouldn't this be `create_ecr_repo_if_needed`?

src/mpyl/utilities/docker/__init__.py
ewaldschrader: done

ewaldschrader:
@Jorg88 @koeves I think it's ready to be merged

src/mpyl/utilities/docker/__init__.py
```diff
@@ -168,6 +171,11 @@ def docker_registry_path(docker_config: DockerRegistryConfig, image_name: str) -
     return "/".join([c for c in path_components if c]).lower()
 
 
+def ecr_repository_path(host_name: str, image_name: str) -> str:
```
koeves, line 174: Can you add unit tests for this method?

src/mpyl/utilities/docker/__init__.py
```diff
@@ -338,3 +350,62 @@ def remove_container(logger: Logger, container: Container) -> None:
...
+        logger.info(f"Repository '{repo}' not found. Creating...")
+        ecr_client.create_repository(
+            repositoryName=repo,
+            imageTagMutability="IMMUTABLE",
+            encryptionConfiguration={
+                "encryptionType": "AES256",
+            },
+        )
+        logger.info(f"Repository '{repo}' created successfully.")
+        ecr_lifecycle_policy(logger, ecr_client, repo)
```
koeves, line 387: These methods cannot throw exceptions?

koeves:
Maybe you can include a quick test report to show how you tested it?

src/mpyl/utilities/docker/__init__.py
ewaldschrader: done

src/mpyl/utilities/docker/__init__.py
ewaldschrader: I've added the possible exceptions

src/mpyl/utilities/docker/__init__.py
```diff
@@ -338,3 +350,72 @@ def remove_container(logger: Logger, container: Container) -> None:
...
+            repositoryNames=[
+                repo.lower(),
+            ]
+        )
+        logger.info(f"Repository '{repo}' exists.")
+    except (
+        ecr_client.exceptions.InvalidParameterException,
+        ecr_client.exceptions.ServerException,
+    ) as exc:
+        raise exc
```
Jorg88, line 380: You can remove the excepts (also on line 396) in which you just raise the exception anyway right? That would happen anyway without the except, or am I missing something?

src/mpyl/utilities/docker/__init__.py
```diff
@@ -338,3 +350,72 @@ def remove_container(logger: Logger, container: Container) -> None:
...
+            ecr_client.exceptions.InvalidTagException,
+            ecr_client.exceptions.TooManyTagsException,
+            ecr_client.exceptions.LimitExceededException,
+        ) as exc:
+            raise exc
+        ecr_lifecycle_policy(logger, ecr_client, repo)
+    logger.info(f"ECR ready, pushing image to {repo}...")
+
+
+def ecr_lifecycle_policy(logger: Logger, ecr_client, repo):
```
Jorg88, line 401: General tip: try to be precise with naming stuff, it might seem insignificant now but it really makes a difference on the long term when maintaining and working on software. If you name a function like this, it doesn't tell the reader what it does - I would guess this one is a data function that gives me the policy.

src/mpyl/utilities/docker/__init__.py
```diff
@@ -338,3 +350,72 @@ def remove_container(logger: Logger, container: Container) -> None:
...
+    lifecycle_policy = {
```
Jorg88, line 402: Did you and Daniel discuss if this should be in mpyl or the config? Basically, if this might need changes or could be repo/company specific, then it's better in the config.

src/mpyl/utilities/docker/__init__.py
```diff
@@ -190,6 +198,10 @@ def push_to_registry(
 
     login(logger=logger, registry_config=docker_config)
     full_image_path = docker_registry_path(docker_config, image_name)
+    repo_path = ecr_repository_path(docker_config.host_name, image_name)
+
+    if docker_config.provider == "aws":
```
Jorg88, line 203: Oh I was thinking, maybe we can at least type the provider as a literal string? https://mypy.readthedocs.io/en/stable/literal_types.html#parameterizing-literals

koeves: Why isn't it an enum?

src/mpyl/utilities/docker/__init__.py
koeves: Yeah I had a comment about the exceptions but I didn't mean to be listed all like this. Was just checking that you put some ecr method calls in a `try ... except` block, but then had other ecr methods in the except block as well (those will not be caught if they throw exceptions). Is that what you had in mind?

src/mpyl/utilities/docker/__init__.py
Jorg88: Also fine, the are on the same page. The frontend guild is moving from enum to string literals, not sure why precisely

src/mpyl/utilities/docker/__init__.py
ewaldschrader: good one. I'll add it to the config

src/mpyl/utilities/docker/__init__.py
ewaldschrader: check, thanks. 
Yes, it adds the lifecycle policy to a newly created repository

src/mpyl/utilities/docker/__init__.py
ewaldschrader: I'm unsure of any other exceptions that could be thrown.

src/mpyl/utilities/docker/__init__.py
ewaldschrader: added enum class Provider

src/mpyl/utilities/docker/__init__.py
ewaldschrader: changed function name

src/mpyl/utilities/docker/__init__.py
ewaldschrader: added a lifecycle policy check exception

koeves (approved):
Looks good now! Do you want to include this in the next release? We would like to release Monday latest

ewaldschrader:
@koeves Yes that sounds good. Just need to fix the linting, then it's all done :)

koeves:
> All done! Can I merge it already or wait til Monday? @koeves

Yes, you can merge it :)

PR #352, opened 2024-01-30
Feature/tech 779 small improvements

Jorg88:


src/mpyl/artifacts/build_artifacts.py
```diff
@@ -132,6 +134,8 @@ def push(
 
                 artifact_repo.stage(".")
                 artifact_repo.commit(message)
+                if remote_branch_exists:
+                    artifact_repo.pull()  # to prevent issues with parallel runs pushing to the same branch
```
koeves, line 138: If we want to take care of this scenario then I think we need some kind of retry mechanism. Because it could still be that the push of one parallel run happens between the pull and push of the other run

src/mpyl/reporting/targets/github.py
```diff
@@ -76,13 +76,13 @@ def __init__(
...
-        update_stategy: GithubUpdateStategy = GithubUpdateStategy.BODY,
+        update_strategy: GithubUpdateStrategy = GithubUpdateStrategy.BODY,
     ):
         self._raw_config = config
         self._config = GithubConfig.from_config(config)
         self.git_repository = Repository(RepoConfig.from_config(config))
         self.compose_function = compose_function
-        self.update_strategy: GithubUpdateStategy = update_stategy
+        self.update_strategy: GithubUpdateStrategy = update_strategy
         self.body_separator = "----"
```
koeves, line 86: Unrelated, but would be good to do it here. It still happens that if the description contains the string `----` it attaches the mpyl part after it (meaning a good part of the description is now lost). Can we think of a smarter way to detect whether an mpyl run has already occurred and the description was already updated once?

koeves: Like checking for the substring
```
----
### 📕
```

koeves (approved):

PR #354, opened 2024-02-05
[EA-186]: Add deploy target to Acceptance and Test

vbrenister:
There is a requirement to add an Jenkins option to deploy to both acceptance and test. 

We need to add that option to `Deploy Target` enum in order to pass validation

Additionally added `VSCode` and `Metals` related `gitignore` exclusions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,6 +132,9 @@ dmypy.json
 .mpyl/
 Project_Default.xml
 
+#VSCode
+.vscode/
+
 #documentation
 docs/*
 !docs/schema
@@ -148,6 +151,9 @@ node_modules/
 #Scala
 .jvmopts
 .bsp/
+
+#Metals
+.metals/
 
 #Customized locally
 mpyl_config.yml
--- a/src/mpyl/schema/run_properties.schema.yml
+++ b/src/mpyl/schema/run_properties.schema.yml
@@ -61,7 +61,7 @@ properties:
           deploy_target:
             description: The deploy target.
             type: ["string", "null"]
-            enum: [ null,  'PullRequest', 'PullRequestBase', 'Acceptance', 'Production' ]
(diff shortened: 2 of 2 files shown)


koeves (changes requested):
I don't think the schema change is necessary here, since from MPyL's perspective we're not deploying to two environments! So to have a valid schema, you could:
- split the `Acceptance & PullRequestBase` string in the jenkinsfile (less preferred imo)
- change the target parameter to be a checkbox per env instead of a list selection. Then you can loop through those options and run mpyl on each target.

Let me know what you think!

vbrenister:
> I don't think the schema change is necessary here, since from MPyL's perspective we're not deploying to two environments! So to have a valid schema, you could:
> 
> * split the `Acceptance & PullRequestBase` string in the jenkinsfile (less preferred imo)
> * change the target parameter to be a checkbox per env instead of a list selection. Then you can loop through those options and run mpyl on each target.
> 
> Let me know what you think!



> I don't think the schema change is necessary here, since from MPyL's perspective we're not deploying to two environments! So to have a valid schema, you could:
> 
> * split the `Acceptance & PullRequestBase` string in the jenkinsfile (less preferred imo)
> * change the target parameter to be a checkbox per env instead of a list selection. Then you can loop through those options and run mpyl on each target.
> 
> Let me know what you think!

Yeah that makes sense. 
I'll dig into second option because it seems more flexible. Thanks for suggestion, I'll close this PR

koeves:
> Yeah that makes sense. I'll dig into second option because it seems more flexible. Thanks for suggestion, I'll close this PR

Nice, let me know if you need help! I think you can just use a `booleanParam` per env btw: https://www.jenkins.io/doc/book/pipeline/syntax/#available-parameters

PR #355, opened 2024-02-06
Feature/dop 1204 implement project ids

Jorg88:


tunaman (approved):

src/mpyl/schema/project.schema.yml
```diff
@@ -35,6 +36,10 @@ properties:
       type: string
     uniqueItems: true
     minItems: 1
+  projectId:
+    description: >-
+      The rancher project id that this project should be deployed to, required for rbac. Represents a domain in rancher.
+    $ref: '#/definitions/dtapValue'
```
koeves, line 42: I would move this to the kubernetes section at least

koeves: Maybe even a specific kubernetes.rancher section

src/mpyl/project.py
```diff
@@ -308,8 +308,18 @@ def from_config(values: dict):
         )
 
 
+@dataclass(frozen=True)
+class Rancher:
+    project_id: str
```
koeves, line 313: Okay, one last comment about this I promise :D 
Why is this mandatory? From an mpyl perspective I mean

Jorg88: It's not really, only if you have `kubernetes.deployment` and it defaults to an empty string.

koeves: Shouldn't we still be able to provide a default project ID per env though? I mean, we used it like that before and now we want to be able to set the projectId per project, but setting the project IDs per environment also makes sense (even though we want to do something else now)

koeves: My question was whether we _need_ to set the project ID for every project if it uses Kubernetes Deploy. From an mpyl's perspective, I don't think the project ID should be mandatory per project

koeves (approved):

PR #356, opened 2024-02-07
[HEM-607] Add suppert for modifying deployment strategy

nik-kashi:
Add support for changing the deployment strategy and its parameters.


src/mpyl/project.py
```diff
@@ -321,9 +321,13 @@ class Kubernetes:
     command: Optional[TargetProperty[str]]
     args: Optional[TargetProperty[str]]
     labels: Optional[list[KeyValueProperty]]
+    strategy_type: str
```
Jorg88, line 324: https://mypy.readthedocs.io/en/stable/literal_types.html#parameterizing-literals or an enum?

src/mpyl/project.py
```diff
@@ -336,6 +340,9 @@ def from_config(values: dict):
             command=TargetProperty.from_config(values.get("command", {})),
             args=TargetProperty.from_config(values.get("args", {})),
             labels=list(map(KeyValueProperty.from_config, values.get("labels", []))),
+            strategy_type=deployment_strategy.get("strategyType", "RollingUpdate"),
+            max_surge=deployment_strategy.get("maxSurge", "25%"),
+            max_unavailable=deployment_strategy.get("maxUnavailable", "25%"),
```
Jorg88, line 345: Maybe combine the 3 new values into a `DeploymentStrategy` object or something? Also in the ChartBuilder? They belong together and it would keep the properties list a bit shorter.

koeves, line 345: I would put the default values in the mpyl_config itself (you can see there are already default values encoded there). Because right now the defaults are still hard coded in this section, but it might be nice to be able to configure those as well. Now that we're changing this anyway :)

koeves: For example take a look at the following dataclass that parses the default values for the resources from the mpyl config `mpyl.steps.deploy.k8s.chart.ResourceDefaults`

ptab:
Nice feature! 👍 

For future reference, here's what this change would look like if we were using Helm charts: https://github.com/Vandebron/argocd/pull/101/files

Jorg88:
> Nice feature! 👍
> 
> For future reference, here's what this change would look like if we were using Helm charts: https://github.com/Vandebron/argocd/pull/101/files

So very similar except that we also have unit tests

ptab:
Come on, at least _try_ to be impartial 😅 

The PR I linked to is roughly **20x smaller** than this one, plus it's [deployed to an actual Kubernetes cluster](https://github.com/Vandebron/argocd/actions/runs/7846533334/job/21413450865) so we are certain it works in a real environment before it's even merged to the main branch. Opinions and preferences aside, the Helm-based change is objectively smaller and safer to merge.

--- 

In any case, since we're not yet discussing whether to adopt Helm or not, I suggest we pause this topic for now. My goal was just to leave a link here so we can compare real-life examples when we do pick up this topic again in the future 🙂

SamTheisens:
@nik-kashi standard k8s manifest snippets can be mixed into `project.yml` like this https://github.com/Vandebron/mpyl/pull/362. It reduces the amount of custom parameters you need to add, the schema is k8s own (so guaranteed to be correct) and people can consult the standard k8s documentation to find out what it does.

Jorg88:
> Come on, at least _try_ to be impartial 😅
> 
> The PR I linked to is roughly **20x smaller** than this one, plus it's [deployed to an actual Kubernetes cluster](https://github.com/Vandebron/argocd/actions/runs/7846533334/job/21413450865) so we are certain it works in a real environment before it's even merged to the main branch. Opinions and preferences aside, the Helm-based change is objectively smaller and safer to merge.
> 
> In any case, since we're not yet discussing whether to adopt Helm or not, I suggest we pause this topic for now. My goal was just to leave a link here so we can compare real-life examples when we do pick up this topic again in the future 🙂

I actually meant it in a positive way.. if you think away the test logic and add `strategy_type` to your pr, the approach is very similar.  🤷

nik-kashi:
> @nik-kashi standard k8s manifest snippets can be mixed into `project.yml` like this #362. It reduces the amount of custom parameters you need to add, the schema is k8s own (so guaranteed to be correct) and people can consult the standard k8s documentation to find out what it does.

Thanks Sam
But what will happen if we want to modify only one field?
I believe it won't work with your implementation
https://github.com/Vandebron/mpyl/pull/362/files#diff-f2e93ed09e3ea127c21228d5487fed442fe1b3364117f6a7ec4705cd660a72a0R767-R769

src/mpyl/project.py
nik-kashi: Addressed in https://github.com/Vandebron/mpyl/pull/356/commits/69e15036831fbe7c27b8f0e4f1c03a77a26522a4

src/mpyl/project.py
nik-kashi: Addressed in https://github.com/Vandebron/mpyl/pull/356/commits/69e15036831fbe7c27b8f0e4f1c03a77a26522a4

src/mpyl/project.py
nik-kashi: Addressed [here](https://github.com/Vandebron/mpyl/pull/356/commits/69e15036831fbe7c27b8f0e4f1c03a77a26522a4)

SamTheisens:
> > @nik-kashi standard k8s manifest snippets can be mixed into `project.yml` like this #362. It reduces the amount of custom parameters you need to add, the schema is k8s own (so guaranteed to be correct) and people can consult the standard k8s documentation to find out what it does.
> 
> Thanks Sam But what will happen if we want to modify only one field? I believe it won't work with your implementation https://github.com/Vandebron/mpyl/pull/362/files#diff-f2e93ed09e3ea127c21228d5487fed442fe1b3364117f6a7ec4705cd660a72a0R767-R769

Ah yeah, correct. I forgot that. You need to merge the two dictionaries like:
```python
merged_config = {**self.config_defaults.deployment_strategy, **(self.deployment_strategy or {})}
strategy = ChartBuilder._to_k8s_model(merged_config, V1DeploymentStrategy)
```

nik-kashi:
In favor of #362

nik-kashi:
Closed in favor of #362

PR #360, opened 2024-02-08
[TECH-704] Not deploy older release than currently deployed

peterrutg:


Jorg88:
Closing since this seems inactive and will be replaced by argocd anyway. You can always re-open it if it becomes relevant again

PR #374, opened 2024-03-06
Add support for using PR-NUMBER interpolation in build args

MatthijsBon:
branch:


src/mpyl/steps/build/docker_build.py
```diff
@@ -133,3 +137,9 @@ def execute(self, step_input: Input) -> Output:
             message=f"Failed to build docker image for {step_input.project.name}",
             produced_artifact=None,
         )
+
+    @staticmethod
+    def substitute_pr_number(original_value: Optional[str], pr_number: Optional[int]):
+        if original_value and pr_number:
+            return original_value.replace("{PR-NUMBER}", str(pr_number))
+        return original_value
```
koeves, line 145: nit, can be done in one line:

```suggestion
        return original_value.replace("{PR-NUMBER}", str(pr_number)) if original_value and pr_number else original_value
```

src/mpyl/steps/build/docker_build.py
```diff
@@ -133,3 +137,9 @@ def execute(self, step_input: Input) -> Output:
             message=f"Failed to build docker image for {step_input.project.name}",
             produced_artifact=None,
         )
+
+    @staticmethod
+    def substitute_pr_number(original_value: Optional[str], pr_number: Optional[int]):
```
koeves, line 142: You might want to combine this method with `src/mpyl/steps/deploy/k8s/resources/traefik.py:38` (extract that method from the ingress creation and create a helper method to substitute PR number and/or the other wildcards). Not sure about where to put this common method though, so if you don't find a good place, maybe it's fine as is.

MatthijsBon:
I rebased on the 1.4 version of the repo to be able to use it in the monorepo:
* ✅ [Used it](https://jenkins.k8s-serv-backend.vdbinfra.nl/view/MPyL/job/MPyL%20Pipeline/view/change-requests/job/PR-13463/33/) on a branch that doesn't use build-args
* ✅ [Used it](https://jenkins.k8s-serv-backend.vdbinfra.nl/view/MPyL/job/MPyL%20Pipeline/view/change-requests/job/PR-13584/7/) on a branch that does use it

image
🎉

src/mpyl/steps/build/docker_build.py
MatthijsBon: Updated! 👍

src/mpyl/steps/build/docker_build.py
MatthijsBon: I had a look, but I honestly wouldn't know where, so I'm leaving it where it is 😁

Jorg88: Since Daniel is out of office this week, I can take a look tomorrow. I'll also fix the lint error then

PR #377, opened 2024-03-14
Feature/guild xxx implement different way to create build plan

Jorg88:
- [x] Rebase and solve merge conflicts with v1.5 since the pr was branched from v1.4.20 in order to be able to use & test it in the monorepo & github actions
- [ ] Test with tag flow

----

 # ⚠️ Could not find ticket corresponding to `feature/GUILD-XXX-improve-git-logic. Does your branch name follow the correct pattern?


mpyl-dagster-example.py
```diff
@@ -72,14 +72,12 @@ def deploy_projects(
...
-        changes_in_branch = repo.changes_in_branch_including_local()
+        changes_in_branch = repo.changes_in_branch_including_unversioned_files()
         project_paths = repo.find_projects()
     all_projects = set(
         map(lambda p: load_project(Path("."), Path(p), strict=False), project_paths)
     )
-    invalidated = find_invalidated_projects_for_stage(
-        all_projects, stage, changes_in_branch
-    )
+    invalidated = build_project_executions(all_projects, stage, changes_in_branch)
```
ptab, line 80: Small improvement: `invalidated` → `project_executions`

src/mpyl/cli/repository.py
```diff
@@ -130,11 +130,10 @@ def status(obj: RepoContext):
                 )
             )
         else:
-            changes = repo.changes_in_branch()
             console.log(
                 Markdown(
                     f"Revision for `{repo.main_origin_branch}` not found. Cannot diff with base. "
-                    f"*{len(changes)}* (grafted) commits on `{repo.get_branch}`"
+                    f"*1* (grafted) commits on `{repo.get_branch}`"
```
ptab, line 136: What does "grafted" mean in this context?
Also, if it cannot diff with the base branch, how would it know how many extra commits there are on the current branch? Let's consider removing this message if it's not adding any value.

src/mpyl/stages/discovery.py
```diff
@@ -27,89 +40,100 @@ def is_invalidated(
...
-        if stage == deploy.STAGE_NAME or output_invalidated(output, history.hash):
-            relevant.update(history.files_touched)
-        else:
-            return relevant
+    return touched_dependency is not None
 
-    return relevant
 
+def is_output_cached(output: Optional[Output], cache_key: str) -> bool:
+    if output is None or not output.success or output.produced_artifact is None:
```
ptab, line 51: Like we discussed earlier today: adding an extra condition here like `or not output.produced_artifact.hash` might be good to avoid having to generate a random key in case there are no changes.

ptab (commented):
Awesome change! I left some comments for us to address tomorrow.

src/mpyl/cli/repository.py
Jorg88: This is part of the `repo status` function that you want to remove anyway, so I won't touch it in this pr.

ptab:
Tested this feature against [this PR in the monorepo](https://github.com/Vandebron/Vandebron/pull/13761) and after some fighting with Jenkins I'm happy to say it works perfectly! 🎉

src/mpyl/cli/build.py
```diff
@@ -241,14 +239,21 @@ def run(
     required=False,
     help="Comma separated list of the projects to build",
 )
+@click.option(
+    "--stage",
+    default=None,
+    type=str,
+    required=False,
+    help="Stage to run",
```
koeves, line 247: Stage to get the status for?

src/mpyl/project_execution.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+@dataclass(frozen=True)
+class ProjectExecution:
+    project: Project
+    cache_key: str
+    cached: bool
+
+    @staticmethod
+    def always_run(project: Project):
```
koeves, line 15: I see this is only used in unit tests.. Can we move them to a test resources helper class? so move to tests/ from src/

src/mpyl/reporting/formatting/markdown.py
```diff
@@ -40,36 +41,36 @@ def __add_link_if_service(name: str, output: Output) -> str:
...
-        for proj in sorted_plan:
-            project_names.append(wrap_project_name(proj, result))
+        sorted_plans = sorted(plan, key=operator.attrgetter("name"))
+        for project_execution in sorted_plans:
+            project_names.append(wrap_project_name(project_execution, result))
     else:
         project_names = list(map(lambda r: f"_{r.project.name}_", result))
 
-    return f'{", ".join(project_names)}'
+    return f'&nbsp;  {", ".join(project_names)}'
```
koeves, line 64: Why do we need this HTML character? Can't we just use normal space?

src/mpyl/stages/discovery.py
```diff
@@ -60,159 +77,139 @@ def __has_invalidated_dependencies(
...
-        return True
-
-    for rev in sorted(change_history, key=lambda c: c.ord):
-        if rev.hash == file_hash:
-            return False
-        if rev.hash == artifact_hash:
-            return True
-    return True
-
+def hashed_changes(files: set[str]) -> str:
```
koeves, line 100: It's a bit weird to me that the hashing happens when we figure out the build plan. Shouldn't we hash when we're done with a stage? Maybe I'm not getting the new approach fully yet :D

src/mpyl/project_execution.py
Jorg88: It's also used for manual selection

src/mpyl/reporting/formatting/markdown.py
Jorg88: If I add normal spaces it just alligns it to the left for some reason

src/mpyl/steps/run.py
```diff
@@ -124,9 +133,7 @@ def results_for_stage(self, stage: Stage) -> list[StepResult]:
...
 
-    def plan_for_stage(self, stage: Stage) -> set[Project]:
-        plan: Optional[set[Project]] = self.run_plan.get(stage)
-        if plan:
-            return plan
+    def plan_for_stage(self, stage: Stage) -> set[ProjectExecution]:
+        plan: Optional[set[ProjectExecution]] = self.run_plan.get(stage)
 
-        return set()
+        return plan if plan else set()
```
koeves, line 139: nit: plan or set() should work as well

src/mpyl/steps/run_properties.py
```diff
@@ -46,7 +48,13 @@ def construct_run_properties(
                 if explain_run_plan:
                     build_set_logger.setLevel("DEBUG")
                 run_plan = _create_run_plan(
```
koeves, line 50: I'd use named parameters here

src/mpyl/utilities/repo/__init__.py
```diff
@@ -133,6 +134,7 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
     @staticmethod
     def from_clone(config: RepoConfig, repo_path: Path):
+        logging.basicConfig(level=logging.DEBUG)
```
koeves, line 137: Why do we need to set this here?

src/mpyl/utilities/repo/__init__.py
```diff
@@ -244,52 +249,58 @@ def changes_between(self, base_revision: str, head_revision: str) -> list[Revisi
...
-        base_ref = self.base_revision
-        base_hex = base_ref.hexsha if base_ref else self.root_commit_hex
+    def changes_in_branch(self) -> Changeset:
+        return Changeset(self.get_sha, self.changed_files_in_branch())
 
-        head_hex = self._repo.active_branch.commit.hexsha
-
-        return self.changes_between(base_hex, head_hex)
+    def changed_files_in_branch(self) -> set[str]:
+        # TODO pass the base_branch as a build parameter, not all branches  # pylint: disable=fixme
```
koeves, line 258: TODO for when? Can we add the ticket number here?

src/mpyl/utilities/repo/__init__.py
```diff
@@ -244,52 +249,58 @@ def changes_between(self, base_revision: str, head_revision: str) -> list[Revisi
...
 
-    def changes_in_branch_including_local(self) -> list[Revision]:
-        in_branch = self.changes_in_branch()
-        in_branch.append(
-            Revision(len(in_branch), self.get_sha, self.changes_in_commit())
+    def changes_in_branch_including_local(self) -> Changeset:
+        return Changeset(
+            sha=self.get_sha,
+            files_touched=(
+                self.changed_files_in_branch() | self.files_changed_locally()
```
koeves, line 279: I had to look up what `|` is doing in python, I'd just use
```suggestion
                self.changed_files_in_branch().union(self.files_changed_locally())
```

src/mpyl/reporting/formatting/markdown.py
koeves: Can you try with `\t`?

src/mpyl/project_execution.py
koeves: ah okay. I would add a docstring to this method then

src/mpyl/stages/discovery.py
Jorg88: We need to hash again to be able to compare to the stored hash in `is_output_cached`. But maybe @ptab can explain this better.

src/mpyl/utilities/repo/__init__.py
Jorg88: I think this might be a leftover from Pedro when he was debugging, will remove it

src/mpyl/utilities/repo/__init__.py
Jorg88: We do this all over the place actually 😅 maybe you can make a new pr changing it everywhere if you prefer `.union`?

src/mpyl/utilities/repo/__init__.py
Jorg88: We don't have a ticket, it's planned for 1.6.1 - Pedro has a list of things we still need to do. Maybe discuss on Slack if we need tickets

src/mpyl/reporting/formatting/markdown.py
Jorg88: Same 😞

src/mpyl/utilities/repo/__init__.py
ptab: We should have tickets, I haven't had the time to create them yet.

src/mpyl/stages/discovery.py
ptab: Every time we trigger a build with mpyl we hash the content of the files modified in the branch or since a previous tag, so we know whether we need to fully re-run the stage or use the results from a previous run.

The set of modified files does not change throughout a run, so it really doesn't matter when we calculate the hash as long as it's before the check @Jorg88 mentioned. We could even take this hashing completely outside MPyL and pass it as a run parameter (@Jorg88 and I are considering this as a future improvement).

ptab (approved):

src/mpyl/stages/discovery.py
koeves: got it. just for readability I would move it to a separate helper module and write specific tests for it (like the hash changes if a new file is added, the hash goes back to what it was before when that new file is not changed anymore, stuff like that)

ptab: Good suggestions 👍

PR #378, opened 2024-03-15
Drop repo init and repo status commands

ptab:


Jorg88:
Don't forget to remove the repo commands from the mpyl github actions that we use to test the pr & tag flows!

koeves (approved):

PR #386, opened 2024-03-22
fix: do not try to hash deleted files

ptab:


tests/stages/test_discovery.py
koeves, line 1: can you add unit test about this deletion scenario? or is it already being tested (if so, then let's make it more explicit)?

koeves (approved):

koeves (approved):

PR #391, opened 2024-03-26
Upped the general dagster memory limit.

LeopoldGabelmann:
For the dbt project, we were running into OOM and agreed within the dagster guild that we can up the memory limit.


kgrunert (approved):
is this the sensible default from now on?

timvandruenen:
> Is platform also aware @timvandruenen ?

Yes!

PR #398, opened 2024-04-09
Feature/tech 894 account for concurrent runs

Jorg88:


mpyl-reporter.py
Jorg88, line 1: The changes to this file also need to be applied to `deployment/pipeline/mpyl/mpyl-post-run.py` in the monorepo after releasing this

src/mpyl/stages/discovery.py
```diff
@@ -204,7 +204,7 @@ def find_build_set(  # pylint: disable=too-many-arguments, too-many-locals
 
     build_set: dict[Stage, set[ProjectExecution]] = {}
 
-    build_set_file = Path(BUILD_ARTIFACTS_FOLDER) / "build_plan"
+    build_set_file = Path(BUILD_ARTIFACTS_FOLDER) / "build_plan.pickle"
```
ptab, line 207: No problem with this change, just a reminder that we'll need to change [the monorepo workflow](https://github.com/Vandebron/Vandebron/blob/1c8c48a6e861ef33e17e4cb99a29989ae9ed19d0/.github/workflows/mpyl-pull-request.yml#L77) to publish this new filename.

mpyl-reporter.py
```diff
@@ -20,55 +20,60 @@
...
-        config = parse_config("mpyl_config.yml")
-        properties = parse_config("run_properties.yml")
-        run_properties = construct_run_properties(config=config, properties=properties)
-
-        commit_check = CommitCheck(config=config, logger=logger)
-        accumulator.add(commit_check.send_report(run_result))
+    config = parse_config("mpyl_config.yml")
+    properties = parse_config("run_properties.yml")
+    run_properties = construct_run_properties(config=config, properties=properties)
+    run_result: RunResult = RunResult(run_properties=run_properties)
```
ptab, line 33: For my understanding: what's the purpose of `ReportAccumulator` now that we've done this change? It looks like we're accumulating all `run_resuilt-*.pickle` files into a single `RunResult`, and then accumulating this again into an `ReportAccumulator`. Is there an opportunity to simplify this setup?

src/mpyl/stages/discovery.py
Jorg88: I wanted to make it more explicit which file this is now that we have the same file in `.json` format as well

mpyl-reporter.py
ptab: Yeah that makes sense, thanks for explaining! 👍

ptab (approved):

koeves (commented):
Changes look good.

So we don't forget, I would use the github run id instead of a random uuid if it's not hard to supply it

koeves (approved):

PR #399, opened 2024-04-10
fix: catch the proper git error to allow retrying failed pushes; more debug logging

ptab:


src/mpyl/stages/discovery.py
```diff
@@ -87,15 +87,55 @@ def is_dependency_touched(
...
+    project: str,
+    stage: str,
+    output: Optional[Output],
+    cache_key: str,
+) -> bool:
+    if stage == deploy.STAGE_NAME:
+        logger.debug(
+            f"Project {project} will execute stage {stage} again because this stage is never cached"
+        )
+        cached = False
```
Jorg88, line 101: Just put this is as the default value instead of setting it in every `if`?

src/mpyl/stages/discovery.py
```diff
@@ -87,15 +87,55 @@ def is_dependency_touched(
...
+    elif output.produced_artifact.hash != cache_key:
+        logger.debug(
+            f"Project {project} will execute stage {stage} again because its content changed since the previous run"
+        )
+        logger.debug(
+            f"Cached content for previous run: {output.produced_artifact.hash}"
+        )
+        logger.debug(f"Cached content for current run:  {cache_key}")
```
Jorg88, line 129: Wouldn't saying 'cached content' be confusing when we log the cache key?

src/mpyl/stages/discovery.py
ptab: My preference would be to just do:
```python
if something:
  return False
elif something else:
  return False
else:
  return True
```

But apparently it's against the project style guide 🤷

src/mpyl/stages/discovery.py
ptab: Fair, I'll find a better wording.

Jorg88 (approved):

PR #403, opened 2024-04-15
refactor: extract RunPlan as its own domain entity

ptab:
Split from https://github.com/Vandebron/mpyl/pull/400 to make reviewing easier.

### DIFF
--- a/src/mpyl/cli/repository.py
+++ b/src/mpyl/cli/repository.py
@@ -1,4 +1,5 @@
 """Commands related to the VCS (git) repository"""
+
 from dataclasses import dataclass
 from pathlib import Path
 
@@ -11,6 +12,7 @@ from . import (
     create_console_logger,
 )
 from ..constants import DEFAULT_CONFIG_FILE_NAME, DEFAULT_RUN_PROPERTIES_FILE_NAME
+from ..run_plan import RunPlan
 from ..steps.run_properties import construct_run_properties
 from ..utilities.pyaml_env import parse_config
 from ..utilities.repo import Repository, RepoConfig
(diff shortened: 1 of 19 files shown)


src/mpyl/cli/repository.py
```diff
@@ -160,10 +162,12 @@ def init(obj: RepoContext, url: str, pull: int, branch: str, pristine: bool):
 
     console.log("Preparing repository for a new run...")
 
-    with Repository.from_shallow_diff_clone(
-        branch.replace("refs/heads/", ""), url, "main", obj.config, Path(".")
-    ) if pristine else Repository(
-        config=RepoConfig.from_config(parse_config(obj.config))
+    with (
```
Jorg88, line 165: I've seen these changes more often in your pr's, just wondering: is this Black or your IDE settings?

src/mpyl/run_plan.py
```diff
@@ -0,0 +1,27 @@
...
+from .project import Stage
+from .project_execution import ProjectExecution
+
+
+@dataclass(frozen=True)
+class RunPlan:
+    plan: dict[Stage, set[ProjectExecution]]
+
+    @staticmethod
+    def empty() -> "RunPlan":
```
Jorg88, line 14: I have not seen this `"RunPlan"` syntax before, what does it do?

Jorg88 (approved):

src/mpyl/cli/repository.py
ptab: I'm not sure. I have IntelliJ configured to run Black on save:
image
With (what I think are) the default settings:
image

Not sure why it ends up with different styling 🤷

src/mpyl/run_plan.py
ptab: Python is not very good at dealing with forward references (i.e. using types that haven't been defined yet), so in order to deal with this we need to quote the type hint to make it a literal and tell the interpreter to resolve it later. You can find this explained in more detail in https://peps.python.org/pep-0484/#forward-references and https://peps.python.org/pep-0484/#the-problem-of-forward-declarations.

PR #405, opened 2024-04-17
refactor: renamed build_plan file to run_plan

ptab:
The name `build_plan` causes a lot of confusion since it clashes with the _Build_ stage and the `Sbt Build` and `Docker Build` steps. For this reason, after aligning all variables (from "build plan" or "build set" to "run plan") we're now also aligning the name of the file produced.


ptab:
Replaced by https://github.com/Vandebron/mpyl/pull/406

PR #409, opened 2024-04-18
feat: do not use the cache if a dependency changed

ptab:


tests/stages/test_discovery.py
```diff
@@ -66,69 +126,95 @@ def test_should_find_invalidated_test_dependencies(self):
...
-            Path("tests/projects/job/deployment/project.yml"),
-            strict=False,
+            )
+            assert len(project_executions) == 1
+            job_execution = next(
+                p for p in project_executions if p.project.name == "job"
+            )
+            assert not job_execution.cached
+
+    def test_stage_with_build_dependency_changed(self):
```
koeves, line 172: Do we execute all stages for the build dependency now?

src/mpyl/stages/discovery.py
```diff
@@ -232,31 +241,28 @@ def build_project_execution(
                     f"Project {project.name}: using hash of modified files as cache key {cache_key}"
                 )
             else:
+                # pylint: disable=fixme
+                # FIXME ptab using revision as a cache_key makes no sense, as we will never check against
+                #  that commit against. Would be simpler just to fallback to the UUID generation here for consistency.
```
Jorg88, line 246:
```suggestion
                #  that commit again. Would be simpler just to fallback to the UUID generation here for consistency.
```

tests/stages/test_discovery.py
```diff
@@ -25,24 +29,80 @@
...
+                spec=DockerImageSpec(image="docker-image-path"),
+                hash=hashed_contents,
+            ),
+        ).write(
+            target_path=path,
+            stage=stage.name,
+        )
+        yield path
+    finally:
+        shutil.rmtree(path)
```
Jorg88, line 60: Can you explain what is happening here? It seems to me you're writing the output to the path and then remove everything in the path again 😅 that would be very weird, so I must be missing something

Jorg88 (approved):
Nice work on the unit tests!

tests/stages/test_discovery.py
ptab: No, only stages that match the dependency stage are executed:
* when running the build stage, trigger projects that are build dependencies
* when running the test stage, trigger projects that are test dependencies
* when running the deploy stage, trigger projects that are deploy dependencies
* and so on

The `test_` prefix might be confusing you here, it doesn't refer to the "test" stage but to the fact that this function is a unit test (and all unit tests start with `test_" prefix) 🙂

koeves: Yeah that's what I thought.. So currently we have 2 bugs in mpyl. One is that the dependencies are not taken into account when caching (which this PR fixes), and the other is that build dependencies only trigger the build stage, not all downstream stages. The second bug was introduced in the configurable stages PR, but it is a serious bug in my opinion, because currently a lot of projects depend on this in the monorepo (that's why it was introduced to mpyl as well).

My idea would be to extend the stages definition to account for this type of dependencies. Normal behaviour would be what we have now, but for build, I would add in the run_properties and in the stages schema something like:

```yaml
stages:
  - name: 'build'
    icon: '🏗️'
    dependency_strategy: 'BuildAll'
  - name: 'test'
    icon: '📋'
    # dependency_strategy: 'BuildStage' (default)
  - name: 'deploy'
    icon: '🚀'
  - name: 'postdeploy'
    icon: '🦺️'
```

ptab: Fully agree with you! The only thing I'd challenge is the categorizing: I don't think this behaviour is per se a bug, it seemed to have been changed deliberately by a previous release. I do agree it breaks some expectations on our side though 👍 

I think your suggestion makes a lot of sense, maybe we can book a session to go over these rules in more detail?

PR #410, opened 2024-04-18
feat: always produce hashed contents if changes are available

ptab:
The current implementation needlessly links the fact that the current run can be cached to the production a hash of the contents, which means in some cases forcing a stage to run regardless of its caching status also prevents the next run from being cached.

This PR fixes that by ensuring that, regardless of whether the current run uses cached contents or re-runs the stage, a hash of the changes is always generated.

### DIFF
--- a/src/mpyl/project_execution.py
+++ b/src/mpyl/project_execution.py
@@ -1,6 +1,5 @@
 """This module contains the ProjectExecution class."""
 
-import uuid
 from dataclasses import dataclass
 from typing import Optional
 
--- a/src/mpyl/project_execution.py
+++ b/src/mpyl/project_execution.py
@@ -1,6 +1,8 @@
 """This module contains the ProjectExecution class."""
+
 import uuid
 from dataclasses import dataclass
+from typing import Optional
 
 from .project import Project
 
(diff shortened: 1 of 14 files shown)


src/mpyl/project_execution.py
```diff
@@ -1,21 +1,33 @@
...
+    @staticmethod
+    def skip(project: Project, hashed_changes: Optional[str] = None):
+        return ProjectExecution(
+            project=project, hashed_changes=hashed_changes, cached=True
+        )
+
+    @staticmethod
+    def run(project: Project, hashed_changes: Optional[str] = None):
         return ProjectExecution(
-            project=project, cache_key=uuid.uuid4().hex, cached=False
```
ptab, line 18: We no longer set the hashed contents to a random UUID because that makes it impossible to detect after the fact whether there were changes in the project or not.

src/mpyl/stages/discovery.py
```diff
@@ -227,39 +221,31 @@ def build_project_execution(
...
-                    f"Project {project.name}: no content changes, falling back to revision as"
-                    f" cache key: {changeset.sha}"
-                )
-                cache_key = changeset.sha
 
-            return ProjectExecution(
+            return ProjectExecution.create(
                 project=project,
-                cache_key=cache_key,
                 cached=is_project_cached_for_stage(
```
ptab, line 241: Ideally we would move this method inside `ProjectExecution`, however that creates a circular reference with `project_execution.py` → `models.py` → `project_execution.py` so we can't easily do that. I might deal with that in a follow-up PR.

src/mpyl/project_execution.py
```diff
@@ -1,21 +1,33 @@
...
 class ProjectExecution:
     project: Project
-    cache_key: str
+    hashed_changes: Optional[str]
     cached: bool
 
     @staticmethod
-    def always_run(project: Project):
-        """Create a ProjectExecution for a project that will always run, regardless of caching"""
+    def create(project: Project, cached: bool, hashed_changes: Optional[str] = None):
```
Jorg88, line 16: nit: this structure feels a little bit weird, I guess we could create a factory pattern but then again that also doesn't really fit with Python maybe

Jorg88 (approved):

src/mpyl/project_execution.py
ptab: I do agree, that is kinda what I meant with [this comment](https://github.com/Vandebron/mpyl/pull/410/files#r1571931240). I can take a shot at making this a bit better in a future PR 🙂

PR #414, opened 2024-04-23
feat: diff any kind of tagged commits

ptab:


src/mpyl/utilities/repo/__init__.py
```diff
@@ -231,32 +228,21 @@ def changes_in_branch_including_local(self) -> Changeset:
...
-        logging.debug(f"Current revision: {current_revision} tag: {curr_rev_tag}")
-
-        if curr_rev_tag != current_tag:
-            logging.error(f"HEAD is at {curr_rev_tag} not at expected `{current_tag}`")
-            return Changeset.empty(self.get_sha)
-
-        return self.changes_in_merge_commit()
+    def changes_in_tagged_commit(self, tag: str) -> Changeset:
+        head_commit = self._repo.head.commit
+        tag_commit = self._repo.tag(f"refs/tags/{tag}").commit
```
Jorg88, line 233: Maybe turn them into a string right away since I think you only use them as strings? (except for the if which will still work anyway I guess)

src/mpyl/utilities/repo/__init__.py
```diff
@@ -231,32 +228,21 @@ def changes_in_branch_including_local(self) -> Changeset:
...
-        logging.debug(f"Parent revisions: {parent_revs}")
-        files_changed = set(
-            self._repo.git.diff(
-                f"{str(parent_revs[0])}..{str(self._repo.head.commit)}",
-                name_status=True,
-            ).splitlines()
-        )
-        return Changeset.from_diff(sha=str(self.get_sha), diff=files_changed)
+
+        diff = set(self._repo.git.diff(f"{tag}^..{tag}", name_status=True).splitlines())
```
Jorg88, line 244: What does `{tag}^..{tag}` do precisely? Seems weird to compare between the same tag

src/mpyl/utilities/repo/__init__.py
ptab: Actually, I just remembered that [f-strings automatically convert objects to strings](https://peps.python.org/pep-0498/#specification) so there's no need to manually do `str(...)` anyway. I'll clean it up 👍

Jorg88 (approved):

PR #426, opened 2024-05-23
Implement a new way of sub-selecting stages and projects

Jorg88:


src/mpyl/steps/deploy/k8s/chart.py
```diff
@@ -703,20 +705,25 @@ def _get_env_vars(self):
             if self.step_input.run_properties.versioning.tag
             else self.step_input.run_properties.versioning.pr_number
         )
+        all_deploy_project_executions: set[ProjectExecution] = next(
+            project_executions
+            for stage, project_executions in self.step_input.run_properties.run_plan.full_plan.items()
```
Jorg88, line 710: Here is the actual pr namespace linkup fix by using the full plan (aka the stored file) which includes all the other projects within the deploy stage that are running in different processes in the action.

src/mpyl/build.py
```diff
@@ -51,59 +52,26 @@ def write_run_plan_as_json():
...
-        )
-        console.print(
-            Markdown(f"**Branch:** `{branch}`. Base {base_revision_specification}. ")
-        )
+    logger.info(f"MPyL log level is set to {run_properties.console.log_level}")
 
     result = RunResult(run_properties=run_properties)
-    if result.has_run_plan_projects:
+    if result.has_run_plan_projects():
         console.print(
```
Jorg88, line 70: This has to stay a print since the logger can't handle a markdown object

koeves: There's `mpyl.utilities.logging.try_parse_ansi` for that!

src/mpyl/run_plan.py
```diff
@@ -1,38 +1,61 @@
...
 from dataclasses import dataclass
+from typing import Optional
 
 from .project import Project, Stage
 from .project_execution import ProjectExecution
 
 
-@dataclass(frozen=True)
+@dataclass()
 class RunPlan:
```
koeves, line 10: Feels a bit weird to have a non frozen "dataclass" that actually has state (with the add_state and update methods). Can't we get rid of those and initialise this dataclass with the full runplan? Then we can make it frozen as well. If it's not possible then I would make this a normal class

src/mpyl/run_plan.py
```diff
@@ -1,38 +1,61 @@
...
+        self._selected_projects = projects
+
+    @property
+    def selected_plan(self) -> dict[Stage, set[ProjectExecution]]:
+        if not self.selected_stage and not self._selected_projects:
+            return self.full_plan
 
-    def items(self):
-        return self.plan.items()
+        self._selected_plan = {}
```
koeves, line 41: Don't think this should be a class variable. I think you can just do `selected_plan = {}`.

src/mpyl/steps/run.py
```diff
@@ -82,17 +81,13 @@ def run_properties(self) -> RunProperties:
...
-            len(project_execution) == 0
-            for stage, project_execution in self.run_plan.items()
-        )
+    def has_run_plan_projects(self, include_cached_projects: bool = True) -> bool:
+        for project_executions in self.run_plan.selected_plan.values():
+            for project_execution in project_executions:
+                if include_cached_projects or not project_execution.cached:
+                    return True
+
+        return False
```
koeves, line 90: I think you can do something like
```suggestion
    def has_run_plan_projects(self, include_cached_projects: bool = True) -> bool:
        return any(
            include_cached_projects or not project_execution.cached
            for project_executions in self.run_plan.selected_plan.values()
            for project_execution in project_executions
        )
```

koeves (approved):

koeves (approved):

PR #431, opened 2024-05-31
cleanup: make console width not configurable, let the tool choose automatically

ptab:
[The setting to configure the console width is not working anyway](https://github.com/Vandebron/Vandebron/pull/14594), so it's easier to just let the underlying console library automatically choose the maximum available line width.


ptab:
It looks like `rich` does not really play well with Github Actions and fails to detect the proper width, meaning this PR is not very useful.

PR #433, opened 2024-06-16
Add docker as generic docker registry host

SamTheisens:
Schema documentation says 'aws' is the default registry provider. The
code now falls back on 'aws' when the provider is unspecified, making
what the documentation says true.


Jorg88 (approved):
Nice to see a pr from you again @SamTheisens !

koeves (approved):
Welcome back :)

PR #434, opened 2024-06-17
fix: avoid empty lines everywhere; also print messages sent to stderr

ptab:
Turn this:
image
Into this:
image


koeves (approved):

SamTheisens:
Nice @ptab 👏 ! I just noticed the same problem.

ptab:
Thanks! It was easy to find since I was the one who introduced the problem in the first place 😅

PR #437, opened 2024-06-20
Feature/g UI ld 986 remove cypress postdeploy support

Jorg88:
This is pretty vandebron specific logic which we're moving to a github action, thus it makes sense to remove it from mpyl.


koeves (commented):
I might have missed something, but why do we need to remove this from mpyl? It doesn't work currently?

Jorg88:
> I might have missed something, but why do we need to remove this from mpyl? It doesn't work currently?

It's working, but we thought it was pretty vandebron specific and it's not the best way of running cypress.

ptab (approved):
image

🤩

PR #438, opened 2024-06-24
fix: revert back to the traefik.containo.us api version

ptab:
Our platform is not ready for this upgrade so we need to go back to the old `apiVersion` value.


Jorg88 (approved):
Thanks!

koeves (approved):

PR #441, opened 2024-07-01
cleanup: remove scalafmt

ptab:
`scalafmt` is now run as part of a specific stage in our pipeline and shouldn't be set directly in MPyL.


project/plugins.sbt
```diff
@@ -1,4 +1,3 @@
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "2.1.0")
 addSbtPlugin("se.marcuslonnberg" % "sbt-docker" % "1.9.0")
-addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.6")
```
andreazubenko-vandebron, line 3: I remember we discussed something similar recently - keeping the plugins here makes for an easier onboarding and stardarization, whereas asking people to add them to their global sbt has minor performance improvements on the pipeline

Was a common ground found? Personally I don't have a strong opinion and I'm fine with both, but I do want us to be consistent 

(if this PR gets approved as-is and scalafmt is removed, then we need to announce it clearly - I think many people use it from inside sbt and would see this as a breaking change)

ptab: Thanks for the feedback! I think your point is valid but you might be confusing the repositories 🙂

We make this plugin available to the Vandebron's repository by configuring it [in its own `project/plugins.sbt` file](https://github.com/Vandebron/Vandebron/blob/master/project/plugins.sbt#L17).

As far as I can tell this configuration I'm removing was only used to enable the plugin [in one of the test services](https://github.com/Vandebron/mpyl/blob/main/build.sbt), which is no longer necessary since we're not calling it explicitly. This PR should not affect the monorepo in any way (other than not calling `scalafmtCheckAll` every time the `Sbt Build` step is executed).

andreazubenko-vandebron: The first line was enough, indeed I'm a dumbass who didn't check the git repository and my comment makes no sense :P

andreazubenko-vandebron (approved):

Jorg88 (approved):
My only thought is if we should maybe make this configurable instead, but I know next to nothing of scala so not sure if it's important.

PR #448, opened 2024-07-10
Add a barebone lint step

Jorg88:
Just a small draft pr to showcase how a new stage (in this case linting) would look.


ptab:
Required for:
* https://github.com/Vandebron/Vandebron/pull/15597

PR #453, opened 2024-07-16
feat: add buildTool field to project.yml so we can leverage different workflows per project type 

ptab:


tests/projects/spark-job/deployment/project-override-second.yml
```diff
@@ -1,6 +1,6 @@
 name: 'sparkJob-second'
 maintainer: ['MPyL']
-mpylVersion: 1.0.9
+mpylVersion: 1.7.0
```
Jorg88, line 3: this field is for the upgrader script btw, it has nothing to do with the real mpyl version

ptab:
Replaced by:
* https://github.com/Vandebron/mpyl/pull/454
* https://github.com/Vandebron/mpyl/pull/455

PR #459, opened 2024-07-23
ci: added release notes for 1.7.0

ptab:


andreazubenko-vandebron (approved):
I think it would be nice to do this together with the actual changes if we remember, but nice changes!

ptab:
I agree, and would even go further to mention that this is a perfect use case for automation: extracting these release notes from [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) should be trivial and reduce a lot of manual work.

andreazubenko-vandebron:
> I agree, unfortunately I keep forgetting it. I would even go further to mention that this is a perfect use case for automation: extracting these release notes from [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) should be trivial and reduce a lot of manual work.

That's already how it works with https://github.com/Vandebron/github-actions/releases , though it's a bit annoying in our setup we have double rows for the actual commit and PR merge

PR #467, opened 2024-09-02
fix: Dagster deployment using correct context

tijmenstor:


src/mpyl/steps/deploy/dagster.py
```diff
@@ -68,7 +68,7 @@ def execute(self, step_input: Input) -> Output:
         properties = step_input.run_properties
         context = get_cluster_config_for_project(
             step_input.run_properties, step_input.project_execution.project
-        ).cluster_env
+        ).context
```
Jorg88, line 71: Ah, I actually noticed this in `gh-mpyl` and found it very weird. Figured it wasn't used yet or something 😅

Jorg88 (approved):

ptab (approved):

PR #480, opened 2024-09-15
Add gradle steps as example of plugin mechanism

SamTheisens:
Add `gradle` build and test steps to demonstrate the use of the plugin mechanism.


SamTheisens:
@ptab this https://github.com/Vandebron/mpyl/pull/480/commits/983dbf88835e476601e48361dc8102ba5f0ca0e2 should fix the currently broken publication of the docs.

SamTheisens:
@ptab thanks for approving https://github.com/Vandebron/mpyl/pull/485 and https://github.com/Vandebron/mpyl/pull/487.
They are stacked on top of _this_ PR. Much appreciated if you could give this one a look.

ptab:
Hi @SamTheisens! I did take a look at this PR and while I don't have any specific concerns I also don't have a lot of insights on the bigger picture (I'm not familiar with the plugin mechanism or its goals), so that's why I didn't specifically mark it with an approval. I can do it if helps speed up development 🙂

ptab (approved):

SamTheisens:
> Hi @SamTheisens! I did take a look at this PR and while I don't have any specific concerns I also don't have a lot of insights on the bigger picture (I'm not familiar with the plugin mechanism or its goals), so that's why I didn't specifically mark it with an approval. I can do it if helps speed up development 🙂

Thanks! The context is that I would like to enable a more stripped down version of MPyL with fewer dependencies. The `Step` implementations that are currently included pull in docker and kubernets libraries.

Developing the step implementations outside of the MPyL library also gives a much quicker roundtrip, because you can build and test steps locally in the repo in which you're using MPyL.

PR #481, opened 2024-09-16
fix: Force ArgoCD to replace Jobs when syncing

tijmenstor:
MPyL inherently deletes and deploys `Job`s in K8s because they modify immutable fields. To ensure same behavior across ArgoCD implementations, we added this annotation.


ptab (approved):
Eventually we should make these annotations configurable per-project, but for now I'm OK with this as it solves an immediate issue.

PR #486, opened 2024-09-24
feat: support new Traefik 3 API version and configuration syntax

ptab:
This PR upgrades MPyL to use Traefik 3.x.

Changes to notice:
* moved from the `traefik.containo.us/v1alpha1` to the `traefik.io/v1alpha1` API version
* moved from the deprecated `ipWhiteList` middleware setting to `ipAllowList`

### DIFF
--- a/.github/workflows/build-package.yml
+++ b/.github/workflows/build-package.yml
@@ -196,7 +196,7 @@ jobs:
 
       - uses: ./.github/setup-kind
         with:
-          traefik-chart-version: 31.1.1
+          traefik-chart-version: 27.0.2
           prometheus-chart-version: 60.3.0
           sealed-secrets-chart-version: 2.7.3
           spark-operator-chart-version: 1.1.25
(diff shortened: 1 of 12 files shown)


tunaman (approved):

.github/workflows/build-package.yml
```diff
@@ -196,7 +196,7 @@ jobs:
 
       - uses: ./.github/setup-kind
         with:
-          traefik-chart-version: 27.0.2
+          traefik-chart-version: 31.1.1
```
timvandruenen, line 199: Isn't it safer to still use chart version 27.0.2, just like we do in our production cluster?
If we upgrade traefik to v3 in prod, we can also bump it here.

timvandruenen (approved):

.github/workflows/build-package.yml
tunaman: I checked it's a minor Traefik update, I want to propose to actually update in scaleway to this version as well.

ptab: Moved it back to `27.0.2`, everything still passes. The good news is now we know these changes work on both Traefik versions 😉

PR #491, opened 2024-10-28
Feature/data 000 prepare dagster deploy for argocd

kgrunert:


src/mpyl/steps/deploy/dagster.py
```diff
@@ -28,6 +28,224 @@
...
+
+        write_chart(
+            chart={},
+            chart_path=values_path,
+            chart_metadata="",
+            values=user_code_deployment,
+        )
+
+        return Output(
+            True, f"Successfully written helm chart manifest to {values_path}"
```
Jorg88, line 94: You might also want to add a `artifact=input_to_artifact`, `Output` is pretty important for logging and debugging

src/mpyl/steps/deploy/dagster.py
```diff
@@ -28,6 +28,224 @@
...
+        )
+
+        self._logger.debug(
+            f"Writing user code manifest with values: {user_code_deployment}"
+        )
+
+        values_path = Path(step_input.project_execution.project.target_path)
+        self._logger.info(f"Writing Helm values to {values_path}")
+
+        write_chart(
```
Jorg88, line 86: Seems like there is quite some overlap with the k8s deploy step, but just too much difference to align them 😅

src/mpyl/steps/deploy/dagster.py
```diff
@@ -28,6 +28,224 @@
...
+            chart={},
+            chart_path=values_path,
+            chart_metadata="",
+            values=user_code_deployment,
+        )
+
+        if not step_input.dry_run:
```
Jorg88, line 181: Everything above from this is a copy of the `Dagster Helm Template` step right? Maybe optimise that a bit to reduce the amount of copied code? Might also make it more clear which parts are shared

src/mpyl/steps/deploy/dagster.py
```diff
@@ -28,6 +28,224 @@
...
+                self.__flatten_result_messages,
+                results[1:],
+                results[0],
+            )
+            if len(results) > 1
+            else results[0]
+        )
+
+    @staticmethod
+    def __flatten_result_messages(acc: Output, curr: Output) -> Output:
```
Jorg88, line 129:
```suggestion
    def __flatten_outputs(acc: Output, curr: Output) -> Output:
```

src/mpyl/steps/deploy/dagster.py
```diff
@@ -28,6 +28,224 @@
...
+                name="Dagster Template",
+                description="Creates a dagster user code helm chart and adds an entry to dagster's K8s ConfigMap",
+                version="0.0.1",
+                stage=STAGE_NAME,
+            ),
+            produced_artifact=ArtifactType.NONE,
+            required_artifact=ArtifactType.DOCKER_IMAGE,
+        )
+
+    def __evaluate_results(self, results: List[Output]):
```
Jorg88, line 117:
```suggestion
    def __combine_outputs(self, results: List[Output]) -> Output:
```

src/mpyl/steps/deploy/dagster.py
kgrunert: Tried extracting big duplications to `DagsterBase`, wasnt too sure about the `rollout_restart` functions, but in theory, can do that too..

kgrunert: [c788197](https://github.com/Vandebron/mpyl/pull/491/commits/c78819718b08d3dc12d6f4e5b77408dc5f06ed15)

heck, why not

Jorg88 (approved):
Nice!

PR #494, opened 2024-11-15
fix: only detect projects that match the full root path

kgrunert:


src/mpyl/stages/discovery.py
```diff
@@ -28,7 +28,7 @@ class DeploySet:
 
 
 def file_belongs_to_project(project: Project, path: str) -> bool:
-    return path.startswith(str(project.root_path))
```
SamTheisens, line 31: I had exactly the same issue 😄 . Maybe We can leave this to Python's path lib https://github.com/Vandebron/mpyl/pull/493/commits/bbfe7290604ce28dc9d3c8eda00fa14a9010ab3e

kgrunert: also fine with me 😄  shall i close this issue then in favour of your PR?

SamTheisens: SGTM 👍 could you approve that PR in that case?

